PR #204, opened 2014-12-30
Us/1233/expose import export

jespino:


superalex:
I think it would be better if when exporting a project the response mimetype would allow you directly to download a file

jespino:
completely agree

jespino:
Fixed!

superalex:
- Email isn't sent when load_dump is executed via celery
 - It would be great adding throttling to project export to avoid DOS

jespino:
Included both changes!

superalex:
Cool, works great!,

I think the importer should generate the project slug instead reading it from the file :)

superalex:
Cool! :+1:

PR #209, opened 2015-01-07
US #1678: Add webhooks to the backend

jespino:


superalex:
Looks great, just three possible improvements:
 - The test action for a webhook is a GET request, I think it should be a POST
 - Actually we only keep the answered data, it would be great saving the sent data and the response
 - It would be great if you could resend a webhook call, something like a post to /webhooks/X/logs/Y/resend

superalex:
Great work!!! :dancers: :dancers: :dancers: :dancers: :dancers: :dancers: :dancers:

PR #214, opened 2015-01-13
Issue #1734: currentPassword is not needed for github users

bameda:
Merge https://github.com/taigaio/taiga-front/pull/243 too


jespino:
:+1:

PR #215, opened 2015-01-13
Issue #1886: Bad request when use project.slug instead project.id in API...

bameda:
... calls


jespino:
Added logs for ensure detect problems related to request with project equals to undefined.

PR #221, opened 2015-01-21
Import API in synchronous mode now returns the new project serialized

superalex:


taiga/export_import/api.py
```diff
@@ -181,8 +182,9 @@ def load_dump(self, request):
             task = tasks.load_project_dump.delay(request.user, dump)
             return Response({"import-id": task.id}, status=status.HTTP_202_ACCEPTED)
 
-        dump_service.dict_to_project(dump, request.user.email)
-        return Response(None, status=status.HTTP_204_NO_CONTENT)
+        project = dump_service.dict_to_project(dump, request.user.email)
+        response_data = ProjectSerializer(project).data
+        return Response(response_data, status=status.HTTP_200_OK)
```
bameda: I think status must be HTTP_201_CREATED

superalex: You are right, it's updated now ;)

PR #224, opened 2015-01-22
Set reply-to to feedback emails

bameda:


jespino:
I like it :heart: I have to test, but looks good.

superalex:
I'm so happy right now... :+1:

PR #228, opened 2015-01-29
Use our get_object_or_404 function instead the django one

bameda:


jespino:
I :heart: it! :smile:

PR #235, opened 2015-02-02
Fix Issue #2072: I can't add comments after the role points are edited

bameda:
depends on https://github.com/taigaio/taiga-front/pull/312


bameda:
:hankey: :hankey: :hankey: :hankey: :hankey: :hankey: :hankey: :hankey: :hankey: :hankey: :hankey:

bameda:
Fixed in https://github.com/taigaio/taiga-back/pull/237

PR #274, opened 2015-03-10
Hide email to not-admin users

jespino:


bameda:
Moved to https://github.com/taigaio/taiga-back/pull/275

PR #287, opened 2015-03-20
Better gitlab support

Shelnutt2:
This patchset add suport for acting on closed or reopen issues in gitlab. These issues will be opened or closed in taiga. This allows for tighter integration with gitlab.


superalex:
Thank you very much @Shelnutt2 :+1: 

Your PR looks good except for the "close" part. Issue status in Taiga are customizable, your problem could not have a "Closed" status or you could have translated it to another language, for example "Cerrado" in spanish.

I think this PR should include the association between gitlab issue status and that it should be customizable from the admin -> Integrations -> gitlab. Probably it would also need another PR for taiga-front.

Regards!

Shelnutt2:
I will see about adding the option to front end for tieing the two together. You are very correct in that, it needs to be dynamic and user definable. I'll push a new patchset when that is complete.

superalex:
Great!,

Thank you very much @Shelnutt2 :)

PR #293, opened 2015-03-30
Fixing uuid generation

superalex:


superalex:
uuid1 is safe...ignoring this one :(

PR #318, opened 2015-04-29
Uncomment the DATABASES dict in settings/local.py.example, and add a...

reduxionist:
default PASSWORD value to it, in order to account for the addition of Postgresql authorization instructions to the setup-development docs (see https://github.com/taigaio/taiga-doc/issues/52)

This commit is required by https://github.com/taigaio/taiga-doc/pull/62

### DIFF
--- a/settings/local.py.example
+++ b/settings/local.py.example
@@ -16,17 +16,17 @@
 
 from .development import *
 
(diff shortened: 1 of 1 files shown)


superalex:
Great! :+1:

PR #330, opened 2015-05-22
Adding by_username endpoint to users API

superalex:


taiga/users/api.py
```diff
@@ -79,6 +79,14 @@ def list(self, request, *args, **kwargs):
...
         return response.Ok(serializer.data)
 
+    @list_route(methods=["GET"])
+    def by_username(self, request):
+        queryset = self.filter_queryset(self.get_queryset())
+        username = request.QUERY_PARAMS.get("username", None)
+        self.object = get_object_or_404(queryset, username=username)
+        serializer = self.get_serializer(self.object)
+        return response.Ok(serializer.data)
+
```
bameda: Why not?

```
@list_route(methods=["GET"])
def by_username(self, request):
    username = request.QUERY_PARAMS.get("username", None)
    user = get_object_or_404(models.User, username=username)
    return self.retrieve(request, pk=user.pk)
```

superalex:
@bameda I made the changes you suggested!

bameda:
Test with users in public projects.

superalex:
Updating the API so that any user (even anonymous) can access any user detail (no private info shown)

PR #337, opened 2015-05-26
Allow users to enter a ref number in the search

artlepool:
Simple change to add the Ref column in to the columns that are searched when the user searches in a project for matching user stories, tasks and issues.

I looked at updating the tests (taiga-back/tests/integration/test-searches.py), but wasn't too sure how to do this given the Ref is generated automatically when the test user stories, tasks, issues are generated. So would need a bit of advice if you'd like me to add a test for this...


superalex:
Great! :+1:

PR #346, opened 2015-06-03
Adding user creation to user profile timeline

superalex:


taiga/timeline/management/commands/rebuild_timeline_for_user_creation.py
```diff
@@ -0,0 +1,95 @@
...
+import gc
+
+class BulkCreator(object):
+    def __init__(self):
+        self.timeline_objects = []
+        self.created = None
+
+    def createElement(self, element):
+        self.timeline_objects.append(element)
+        if len(self.timeline_objects) > 1000:
```
jespino: What happens when the timeline_objects length is less than 1000? For example, I insert 3 timeline entries, this is never persisted?

superalex: Ops!, I have to fix this part! :dancer:

superalex: Fixed now!

jespino:
The tests are failing because you hasn't the database migration for remove the not-null project foreign key restriction.

superalex:
@jespino you are right, I forgot adding it :(

PR #355, opened 2015-06-17
Supporting a "field type" property for custom fields (taiga-front #451)

artlepool:
This is the taiga-back change required to support the work for taiga-front issue #451, to implement a multi-line text box custom field type.


superalex:
Great! :+1:

PR #361, opened 2015-06-20
Fix import and deletion of projects

jespino:
Some projects import and deletion doesn't work correctly because the events app signal is there, I have fixed this, removing the signal handling from import and deletion (now works).

I have refactored a little the connect/discconect signals. In the past when you delete a project, some signals are disconnected and not reconnected, this is fixed too in this commit.

### DIFF
--- a/taiga/projects/api.py
+++ b/taiga/projects/api.py
@@ -255,30 +255,36 @@ class ProjectViewSet(HistoryResourceMixin, ModelCrudViewSet):
         super().pre_save(obj)
 
     def destroy(self, request, *args, **kwargs):
+        from taiga.events.apps import connect_events_signals, disconnect_events_signals
+        from taiga.projects.tasks.apps import connect_tasks_signals, disconnect_tasks_signals
+        from taiga.projects.userstories.apps import connect_userstories_signals, disconnect_userstories_signals
+        from taiga.projects.issues.apps import connect_issues_signals, disconnect_issues_signals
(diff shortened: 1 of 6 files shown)


superalex:
@jespino tests are failing...could you check this one, please?

jespino:
Ok, I review this :smile:

jespino:
I launch again the tests, and now is working, probably a strange race condition in travis.

taiga/projects/api.py
```diff
@@ -255,31 +255,44 @@ def pre_save(self, obj):
...
+            connect_userstories_signals()
+            connect_memberships_signals()
+            raise e
+
+        connect_events_signals()
+        connect_issues_signals()
+        connect_tasks_signals()
+        connect_userstories_signals()
+        connect_memberships_signals()
+
```
superalex: Why not using a "finally" instead of repeating the code?

jespino:
Probably is the correct way, but I have to test if finally block runs when reraise the exception in the except block. I going to check it.

jespino:
Ok, finally works in the correct way, i going to change it to use finally. Thanks :kissing_heart:

jespino:
Fixed and rebased

superalex:
I've relaunched the tests two times but still failing (something related to memberships), could you please check it?

jespino:
Ok, i will check it again.

superalex:
Works great! :+1:

PR #364, opened 2015-06-23
Issue#2943: Regenerate refs for tasks, issues and user stories on project change

jespino:


superalex:
It's not working properly for issues or tasks. I miss tests for those.

jespino:
Ok, I review this. Thanks :smile:

jespino:
I think now is ready (waiting for the tests). Added some extra functionality like try to detect the status, priority, severity, type of the objects, and remove milestone and user_story, because doesn't be the same in other project.

Need a good review, anyway the code must affect only to the updates where the project_id change.

superalex:
I've been playing with curl and works great! :)

PR #374, opened 2015-06-25
Issue#2572: On status open/closed change recalc the is_closed for use…

jespino:
…r stories

### DIFF
--- a/taiga/projects/apps.py
+++ b/taiga/projects/apps.py
@@ -53,6 +53,18 @@ def connect_projects_signals():
                                  dispatch_uid="update_project_tags_when_create_or_edit_taggable_item_projects")
 
 
+def connect_us_status_signals():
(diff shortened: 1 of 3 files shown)


superalex:
How could I test this one?, I think there was a situation where it was failing, isn't it?

jespino:
The problem was that when you change the is_closed attribute of a TaskStatus or UserStoryStatus, the state of the user stories isn't recalculated. For example, you have a task in "rejected" which is not closed state, and all the other tasks in Done. Then you think, "rejected" must be closed state, going to the admin, change the is_closed of the "rejected" state. Then you expect to have the US with this task closed, but there is the bug.

The fix is simple, when change the UserStoryStatus or TaskStatus, recalc the is_closed of the involved user stories.

superalex:
:+1:  :+1:  :+1:  :+1:  :+1:  :+1:  :+1:  :+1:

PR #383, opened 2015-06-30
Uploading file with multiple extensions

al42and:
When uploading file with multiple dots in its name, like `archive.tar.gz`, only last extension is preserved, and the rest of the filename is passed to `django.template.defaultfilters.slugify` function which removes any dots. As a result, file is saved as `archivetar.gz`.

[Relevant code](https://github.com/taigaio/taiga-back/blob//9d94de9b219c2783673a18d517b36b33a222174c/taiga/projects/attachments/models.py#L35).

Better solution seems to be `django.utils.text.get_valid_filename` function, specifically designed for filtering filenames. This also allows for unicode filenames.
The proposed patch was tested with `stable` and `master` branches of Taiga, installed with taiga-scripts and using Django 1.7.8.

BTW, filename is converted to lowercase, which seems unnecessary to me, but whatever.


jespino:
I like it! :+1:

jespino:
I will remove the .lower() too.

jespino:
Merged manually through cherry-pick. Thanks! :sparkles:

jespino:
And removed the .lower() in master too. Thanks again! :smiley:

al42and:
You also need to remove `.lower()` in [api.py](https://github.com/taigaio/taiga-back/blob/master/taiga/projects/attachments/api.py#L59), otherwise the name in file list in the bottom of wiki page is still lowercase.

jespino:
You are completely right, and in other 2 places (export_import/service.py and sample_data.py). Thanks one more time!

jespino:
Done! :smiley:

al42and:
You're welcome. I'm glad that I was able to help.

PR #412, opened 2015-07-30
Adding deleted datetime to webhook sent info when deleting elements

superalex:


jespino:
I like it, I will test it later :smile:

superalex:
Thanks! :)

PR #427, opened 2015-08-08
Fix membership timeline entries in the user profile

jespino:


superalex:
Great! :+1: 

Thank you very much @jespino

PR #433, opened 2015-08-12
Added notification threading.

### LINKED ISSUE #409: Implement message-id in email notifications so clients can properly thread notifications
Currently, email clients like Thunderbird and Outlook cannot thread notifications about the same item. Adding message-id headers will allow this. Taiga can be noisy with notifications, and threaded messages will help make them more manageable and less daunting.

brettp:
Fixes #409.


taiga/projects/notifications/services.py
```diff
@@ -270,11 +273,24 @@ def send_sync_notifications(notification_id):
     model = get_model_from_key(notification.key)
     template_name = _resolve_template_name(model, change_type=notification.history_type)
     email = _make_template_mail(template_name)
+    site = get_current()
+    domain = site.domain.split(":")[0] or site.domain
```
superalex: In other parts of the application we are reading site and domain from the app settings, settings.SITES["api"]["scheme"], settings.SITES["api"]["domain"]

superalex:
Besides my comment about the settings I really like the PR and think it's really useful! :+1: 

Could you please update the code so we can accept the PR?

taiga/projects/notifications/services.py
```diff
@@ -270,11 +273,24 @@ def send_sync_notifications(notification_id):
...
+    if "ref" in obj.snapshot:
+        id = obj.snapshot["ref"]
+    elif "slug" in obj.snapshot:
+        id = obj.snapshot["slug"]
+
```
superalex: I think a default allways existing value for id should be included

brettp:
@superalex: I found a problem with Outlook so would want to fix that first, too. I'll take care of the issues you mentioned while fixing that.

taiga/projects/notifications/services.py
```diff
@@ -270,11 +273,24 @@ def send_sync_notifications(notification_id):
...
+
+    headers = {"Message-ID": "<%s/%s/%s@%s>" % (notification.project.slug, id, time.time(), domain),
+               "In-Reply-To": "<%s/%s@%s>" % (notification.project.slug, id, domain),
+               "References": "<%s/%s@%s>" % (notification.project.slug, id, domain),
+
+               "List-ID": 'Taiga/%s <taiga.%s@%s>'  % (notification.project.name, notification.project.slug, domain)}
```
bameda: Please, use str.format method instead of % ([PEP 3101](https://www.python.org/dev/peps/pep-3101/))

brettp:
I've updated the PR with the requested and changes, and attempted to fix the Outlook problems. Outlook doesn't use standard headers to thread messages, but I've added experimental support for their base64 encoded binary headers.

superalex:
It looks great but could you please add some tests for the new parts?

Regards!,

brettp:
@superalex I've added tests and cleaned up the MS Thread-ID function to make it less newbie-y. Let me know if you want to see any other changes!

tests/integration/test_notifications.py
```diff
@@ -33,6 +37,7 @@
 from taiga.projects.userstories.serializers import UserStorySerializer
 from taiga.projects.tasks.serializers import TaskSerializer
 from taiga.permissions.permissions import MEMBERS_PERMISSIONS
+from taiga.projects.notifications import services as notification_services
```
superalex: This one is a duplicated, taiga.projects.notificationsservices is imported in line 31

taiga/projects/notifications/services.py
```diff
@@ -356,3 +381,33 @@ def remove_watcher(obj, user):
...
+    6 bytes = First 6 significant bytes of the FILETIME stamp
+    16 bytes = GUID (we're using a md5 hash of the message id)
+
+    See http://www.meridiandiscovery.com/how-to/e-mail-conversation-index-metadata-computer-forensics/
+    """
+
+    import base64
+    import hashlib
+    import struct
+
```
superalex: Why not moving these imports to the top of the file?

superalex:
Thank you very much @brettp 

I was checking your PR with @bameda and we only have two minor comments :+1:

brettp:
@superalex and @bameda: Thanks! I've cleaned up the imports.

superalex:
Great!, thank you very much @brettp

PR #434, opened 2015-08-14
WIP: Improved GitLab hook comments.

brettp:
I've improved the GitLab integration when changing statuses so the comments display something useful. Let me know if this is a direction you'd like to move forward on and I'll refactor the other hooks, too.

### DIFF
--- a/taiga/hooks/bitbucket/api.py
+++ b/taiga/hooks/bitbucket/api.py
@@ -59,13 +59,5 @@ class BitBucketViewSet(BaseWebhookApiViewSet):
 
         return project_secret == secret_key
 
-    def _get_project(self, request):
-        project_id = request.GET.get("project", None)
-        try:
-            project = Project.objects.get(id=project_id)
-            return project
-        except Project.DoesNotExist:
-            return None
-
     def _get_event_name(self, request):
         return request.META.get('HTTP_X_EVENT_KEY', None)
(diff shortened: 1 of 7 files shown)


superalex:
Hello @brettp 

Thank you very much for the PR, sorry for the delay on answering.

I think the PR looks great but, could you include some tests for the new feature?, I think the other hooks could be refactored in the same way too :+1: 

Regards!,

brettp:
I agree about refactoring the other hooks. Is there a reason most of this logic isn't in a parent class? Unless I'm missing something, that'd DRY up the implementation.

superalex:
I think that we had to separate the logic of this parts because it was different but if you find a way of moving common logic to a parent class that would be great! :+1:

brettp:
I've committed an approach and would like some feedback. I focused on the GitLab hooks, and haven't touched the other ones yet so tests will fail. My goals were to:

* DRY up implementation by pulling similar logic into parent classes.
* Standardize and improve the comments generated across all the hooks.
* Use local Taiga accounts when available or mapped.

schmitzhermes:
any updates on this? Or is there any other solution? 
I'd really enjoy having useful information about the commit added to the Taiga issue :-)

bameda:
Hi all, we have been doing some housecleaning work prior to Taiga6 upcoming release and we are rejecting old PRs that we didn't correctly attend back them. Since Taiga6 bring a signfiicant UX/UI reword, we encourage you to wait and see if Taiga6 satisfies this need and propose a new PR if that's not the case. This time we will make sure to review it when it comes through. Thank you very much on behalf of the Taiga Team!

PR #453, opened 2015-09-09
Add filter by email domain on register

jespino:


bameda:
No test

jespino:
Now have tests!

superalex:
I've been thinking about this one and in my opinion we should also control the user email update and limit it to the valid domains too. If we add this one I would rename REGISTER_ALLOWED_DOMAINS to USER_EMAIL_ALLOWED_DOMAINS.

@bameda @jespino what do you think about it?

bameda:
It makes sense. User must use always a valid domain too in:
- register form
- editing their profiles 
- **sending invitations**

PR #459, opened 2015-09-14
Add email field to UserSerializer

astagi:


bameda:
Hi, @astagi 

Why do you need the user email?? This is a deliberate decision because we don't want to expose all the emails via the API. So we created the UserAdminSerializer use for generate the login response.

astagi:
Hi @bameda! yes, I could imagine that :D we're using Taiga with our internal system and we need the email field to recognize the user in our system (we fetch infos from Taiga and we match the user associated to that user only using the email).. any ideas to solve this problem?  Maybe I can update the PR with something like "show the user email only if you're an administrator"?

bameda:
Ummm, you can edit ```taiga.users.api.UsersViewSet.get_serializer_class()``` method and ```return self.admin_serializer_class``` ```if not self.request.user.is_anonymous() and self.request.user.is_superuser``` or something like this.

Some tests about that will be welcome too ;-)

astagi:
Yes get_serializer_class override seems the best option for me, and yes, we just have UserAdminSerializer class, you're right :) working on, thanks!

astagi:
Done @bameda :smiley:

bameda:
Continue in https://github.com/taigaio/taiga-back/pull/460 I added tests to check requests from anonymous users

astagi:
Nice, thanks @bameda :smile:

PR #462, opened 2015-09-15
Add sha1 field to Attachments

astagi:


astagi:
This PR adds "sha1" field to the attachments. We need to know this value, for example we have an external tool to create and update issues inside Taiga, and during the update we need to know if we already uploaded that particular file.

I just erase the sha1 value when the object is saved without a file, but I don't recalculate sha1 if the associated file changes (I think that this something that can't happen, every time an user upload a file a new attachment gets created).

astagi:
Ok performances are not a problem, I added another commit to calculate sha-1 when the Attachment gets saved. I tried with 1GB file and it takes 3 seconds to calculate sha-1.

taiga/projects/attachments/migrations/0006_auto_20150915_1541.py
```diff
@@ -0,0 +1,26 @@
...
+
+from taiga.projects.attachments.models import get_sha1_file
+
+
+def calculate_sha1(apps, schema_editor):
+    Attachment = apps.get_model("attachments", "Attachment")
+    for attachment in Attachment.objects.all():
+        if attachment.attached_file and not attachment.sha1:
+            attachment.sha1 = get_sha1_file(attachment.attached_file.file)
+            attachment.attached_file.file.close()
```
bameda: Yo close the file here after call get_sha1_file and not in the model Attachment.save() Why?

astagi: I got "too many files open" error if I didn't close the file :)

astagi: Long response, sorry :smile: : it looks like overrided save and other things like custom method doesn't get called/exist using the model in that way... I tried with "print" calls, no way! 

Do you think that we should close the file during save? I think that Django provides some mechanism to do ti automagically at runtime...

bameda: No, it's right for me if django makes magic :tophat:.

astagi: It should :) when you use FileField you get an opened file (as you can see there's no "open" call when I calculate the sha1) I suppose that Django provides an automatic file closing process.. it needs some investingation in the Django's source code anyway :) I'll give it a look!

bameda:
We think this is a very good idea but we have two appraisals:

- With the migration ```attachments/migrations/0006_auto_20150915_1541.py``` the (¿db?) server could be very busy during a long long time if the instance has a lot of files (like taiga.io for example), so we think a command to regenerate all the hashs can be better. 

- The hash is calculated every time a user edit an attachment and not necessarily has to update the file. So we can calculate the hash in the pre_save of the viewset class or in a pre_save signal (#$#!!@$#! I don't like model signals) of the model only when file attr is modified.

What do you think?

astagi:
Ok for using a command and ok for signals :) to avoid signals, in save(), we could get the model instance before the super save call and compare the old "attachment" field with the newer one or if you want to avoid to hit the db again we could keep a reference to the attachment attribute in another field in the model class, I used this hack some time ago:

    class MyModel (Model):
        def __init__(*args, **kwargs):
            super(MyModel, self).__init__(*args, **kwargs)
            self._orig_attr = self.attr

          def save(self, *args, **kwargs):
            if self.attr != self._orig_attr:
              # changed, do stuff!
            super(MyModel, self).save(*args, **kwargs)
            self._orig_attr = self.attr

The problem is: how to know if the file attribute is changed without calculating sha1? :) I haven't investigate FileField deeply yet..

astagi:
Sorry @bameda maybe I asked a silly question! :) anytime the file changes another random path gets generated, right? In this case here's how to know if a new file has been changed!

bameda:
yep, you can use this approach or even pre_save singnals:
```
def pre_save_attachments(self, sender, instance, ...):
    if instance.id:
        old_instance = sender.objects.get(id=instance.id)
        if old_instance.file.path == instance.file.path:
             return 

    calculate_hash()
```

You choose the one you prefer.

astagi:
I think to go for the first approach for a little thing like this :) I forgot to mention the pre_save signal but yeah, it is another case of another hit to the database required (when you get the old_instance). 

Also I had a look at the Django implementation of FileField and it looks like it automatically open the file when you get it but it closes the file *only* when you call delete on the instance (https://github.com/django/django/blob/5153a3bfdcec82324d67ff79862384288cf6afe6/django/db/models/fields/files.py)

It's time for me to move some code then, @bameda :smile:

bameda:
Great!! :+1:

astagi:
@bameda changes are ready :smile:

bameda:
https://github.com/taigaio/taiga-back/pull/470

PR #464, opened 2015-09-17
Making total_points and total_sprints nullable values

superalex:


taiga/projects/services/stats.py
```diff
@@ -23,6 +23,18 @@
...
+        and project.total_story_points\
+        or sum(project.calculated_points["defined"].values())
+
+
+def _get_total_milestones(project):
+    return project.total_milestones is not None\
+        and project.total_milestones\
+        or project.milestones.count()
+
+
```
bameda: I don't like this code very much because it can give wrong results when ```project.total_story_points``` has a false boolean value, I prefer to use ```[on_true] if [expression] else [on_false]```:
```
def _get_total_story_points(project):
    return (project.total_story_points if project.total_story_points is not None else
                sum(project.calculated_points["defined"].values()))

def _get_total_milestones(project):
    return (project.total_milestones if project.total_milestones is not None else 
                project.milestones.count())
```

because I think is more idiomatic.

More info at https://docs.python.org/3.5/faq/programming.html#is-there-an-equivalent-of-c-s-ternary-operator.

superalex:
You are right!, I've made the changes you suggested :+1:

PR #479, opened 2015-10-07
Make taiga compatible with python 3.5, migrate to django 1.8 and update the rest of requirements

bameda:


taiga/projects/notifications/models.py
```diff
@@ -61,10 +61,10 @@ class HistoryChangeNotification(models.Model):
                                             verbose_name=_("created date time"))
     updated_datetime = models.DateTimeField(null=False, blank=False, auto_now_add=True,
                                             verbose_name=_("updated date time"))
-    history_entries = models.ManyToManyField("history.HistoryEntry", null=True, blank=True,
+    history_entries = models.ManyToManyField("history.HistoryEntry",
                                              verbose_name=_("history entries"),
                                              related_name="+")
-    notify_users = models.ManyToManyField("users.User", null=True, blank=True,
+    notify_users = models.ManyToManyField("users.User",
```
superalex: why making these changes?, they generate a migration that seems to do nothing...

bameda: According to https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.ManyToManyField.swappable

> null has no effect since there is no way to require a relationship at the database level.

superalex: Ok!, and the migration?, is it really doing anything?

bameda: It was generated automatically by django, so django needs it

superalex: Ok!

Issue #481, opened 2015-10-13
Emoji extension uses wrong path


csfeathers:
Emojis weren't showing for us. It would try to load the PNGs from `localhost`.

I suspect that [this line](https://github.com/taigaio/taiga-back/blob/9226913caab94b2f5e2a761d8e56b613198a656d/taiga/mdrender/extensions/emojify.py#L38) in `taiga/mdrender/extensions/emojify.py` uses the wrong path.

As far as I can tell it should use `STATIC_ROOT` (or `STATIC_URL`?) instead of the `api` stuff.

Note that this is just educated guesswork, I'm not super familiar with the Taiga code.

bameda:
You are right. I've created a PR https://github.com/taigaio/taiga-back/pull/483

PR #483, opened 2015-10-13
Fix issue #3321: Emoji extension uses wrong path

bameda:
Related to https://github.com/taigaio/taiga-back/issues/481


superalex:
Great!, thank you very much @bameda

csfeathers:
Many thanks!

PR #491, opened 2015-10-17
Add "ref" param to resolver API

### LINKED ISSUE #489: API - Resolve resource allow generic "item" param
Discussed here.
https://groups.google.com/forum/#!searchin/taigaio/hubot/taigaio/h1oJmId4V8Q/dkqqPgJYDwAJ

David Barragán said
> Maybe we can improve resolve resource. For example we can add the attribute "item" to find in everi types:
resolve?project=myproject&us=5
{"project": 68815, "us": 291626}
is the same as 
resolve?project=myproject&item=5
{"project": 68815, "us": 291626}
So if you know the ref. but not the type
resolve?project=myproject&item=15
{"project": 68815, "issue": 291637}

If people think this is a good idea I could try implementing it. If would reduce the number of api calls needed to look up a user story, issue, or task by reference Id. It wouldn't break backwards compatibility since it just adds a new optional parameter.

bufke:
Fixes #489

Added ref GET param to resolver API.
Useful if we know the the ref ID but not the type of object it is.
Reduces number of API calls in cases when we want to reference an object
by it's ref ID.
Modified integration test to confirm ref works just like using a us, ect
params and has same permissions.

### DIFF
--- a/taiga/projects/references/api.py
+++ b/taiga/projects/references/api.py
@@ -59,4 +59,18 @@ class ResolverViewSet(viewsets.ViewSet):
             result["wikipage"] = get_object_or_404(project.wiki_pages.all(),
                                                    slug=data["wikipage"]).pk
 
+        if data["ref"]:
+            if user_has_perm(request.user, "view_us", project):
+                us = project.user_stories.filter(ref=data["ref"]).first()
+                if us:
+                    result["us"] = us.pk
+            if user_has_perm(request.user, "view_tasks", project):
+                task = project.tasks.filter(ref=data["ref"]).first()
+                if task:
+                    result["task"] = task.pk
+            if user_has_perm(request.user, "view_issues", project):
+                issue = project.issues.filter(ref=data["ref"]).first()
+                if issue:
+                    result["issue"] = issue.pk
+
         return response.Ok(result)
--- a/taiga/projects/references/api.py
+++ b/taiga/projects/references/api.py
@@ -60,15 +60,18 @@ class ResolverViewSet(viewsets.ViewSet):
                                                    slug=data["wikipage"]).pk
 
         if data["ref"]:
+            ref_found = False  # No need to continue once one ref is found
             if user_has_perm(request.user, "view_us", project):
                 us = project.user_stories.filter(ref=data["ref"]).first()
                 if us:
                     result["us"] = us.pk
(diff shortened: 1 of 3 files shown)


taiga/projects/references/api.py
```diff
@@ -59,4 +59,18 @@ def list(self, request, **kwargs):
...
+                    result["us"] = us.pk
+            if user_has_perm(request.user, "view_tasks", project):
+                task = project.tasks.filter(ref=data["ref"]).first()
+                if task:
+                    result["task"] = task.pk
+            if user_has_perm(request.user, "view_issues", project):
+                issue = project.issues.filter(ref=data["ref"]).first()
+                if issue:
+                    result["issue"] = issue.pk
+
```
bameda: there is only one user story OR task OR issue with a ref so if I've found a story, for example, I don't need to search anymore

bufke:
Added logic to stop searching once a ref is found.

bameda:
- [ ]  After merged remember to update the documentation

bufke:
I could update the docs as part of this pull request if you want.

bameda:
Don't worry. I can do that. Is only that I'm thinking about what happened if issue #55 and #47 exist and I make a request like api/vi/resolve?project=projectexample&issue=55&ref=47? I'm thinking about changing the serializer validator to prevent calls with ref and (us or issue or task params). On the other hand we are engaged preparing the next release,  I hope for next week, but this PR will be.

superalex:
@bameda did you finally make that change in the serializer validator?, is this one ready to test and merge?

bameda:
No, I was trying to fix https://github.com/taigaio/taiga-back/pull/492 because It doesn't work properly

superalex:
Let me fix #492 I've seen some tests are failing, what did you find it isn't working?

bameda:
No, right now tests work fine. But the serializers don't tell me information about if I watch, like or vote an object properly. 

Test your branch with from if you want to see the error.

superalex:
Sorry but I don't fully understand what to test exactly and how...

bameda:
give me 30 minutes...

superalex:
Okidoki! as many as you want :dancer:

bameda:
https://github.com/taigaio/taiga-back/issues/497

PR #496, opened 2015-10-26
Issue 3372: Velocity not calculated properly

superalex:


bameda:
Albondiga!! :dancers:

PR #497, opened 2015-10-26
Added ref GET param to resolver API.

bameda:
Useful if we know the the ref ID but not the type of object it is.
Reduces number of API calls in cases when we want to reference an object
by it's ref ID.
Modified integration test to confirm ref works just like using a us, ect
params and has same permissions.

Related to https://github.com/taigaio/taiga-back/pull/491 with tests and params validator


bameda:
thanks you very mutch @bufke :dancer:

PR #498, opened 2015-10-27
a patch to fix TypeError: sequence item 0: expected str instance, NoneType found

luanntvn:


bameda:
@luanntvn when does it happen? we have never had a bug about this.

luanntvn:
A user reported that he can't assign an issue for someone and I got this error in the log.
I thought I should fix the function lists_diff in taiga-back/taiga/projects/history/templatetags/functions.py instead:
def lists_diff(list1, list2):
    return (list(set(list1) - set(list2)))

Looks like when the list1 contains None items, then (list(set(list1) - set(list2))) will contain None items and caused the error.

Here is my test code:
list1 = [None, 'One', 'Two']
list2 = ['One']
values_removed = list(set(list1) - set(list2))
print (values_removed)
print ('join: ' + ', '.join(values_removed))

bameda:
@luanntvn I try this:

```python
>>> list1 = ["a", None, "b", "c"]
>>> list2 = [None, "b", "d", "e"]
>>> list(filter(None.__ne__, set(list1) - set(list2)))
['c', 'a']
```

So... you can do this

````python
@library.global_function
def lists_diff(list1, list2):
    return list(filter(None.__ne__, set(list1) - set(list2)))
```

what do you think?

luanntvn:
Great! So don't need to merge my old code.

bameda:
Done: https://github.com/taigaio/taiga-back/pull/514

PR #501, opened 2015-10-29
Fix notification level names

bameda:


superalex:
Albóndigas for @bameda !!!! :dancers: :dancers: :dancers: :dancers: :dancers:

PR #511, opened 2015-11-04
Setting finished date correctly on tasks

superalex:


bameda:
Merged and backoported to stable :horse: :pineapple:

PR #515, opened 2015-11-06
Update README.md

andela-ijubril:
Change trasnlate to translate in README file


bameda:
Thanks @andela-ijubril :smile_cat:

PR #528, opened 2015-11-16
Enabling lowercase for username or email when allowed

superalex:


niwinz:
Maybe it should not be merged because more simple solution is possible... (just using some browser input field config).

superalex:
We are actually showing the user info about not using uppercase but there are some browsers prefilling that data with the first letter in uppercase. This tries to solve the case where a user has ignored that alerts and we have just one user with the lower case version of the username or email.

niwinz:
The prefilling data with upper-case case can be solved directly in browser I guess: http://stackoverflow.com/questions/5171764/how-do-you-turn-off-auto-capitalisation-in-html-form-fields-in-ios

I just suggesting that is not necessary touch backend code for solve user interface problems.

Xaviju:
Not sure if the problem is related only to UI only, it is also UX. 
Users write their mail address in both uppercase and lowercase and its accepted always as lowercase in other platforms. They expect their emails to be accepted always regardless of case style. And they don't care about the spec or the database. I would add both fixes.

superalex:
Thanks @niwinz I would add the both fixes as @Xaviju says :+1:

niwinz:
:+1:

Xaviju:
:tada:

taiga/users/services.py
```diff
@@ -47,8 +47,13 @@ def get_and_validate_user(*, username:str, password:str) -> bool:
...
 
     user_model = apps.get_model("users", "User")
-    qs = user_model.objects.filter(Q(username=username) |
-                                   Q(email=username))
+    qs = user_model.objects.filter(Q(username__iexact=username) |
+                                   Q(email__iexact=username))
+    
+    if len(qs) > 1:
+        qs = user_model.objects.filter(Q(username=username) |
+                                       Q(email=username))
```
bameda: why don't you use ```qs``` instead of make a new query?

taiga/users/api.py
```diff
@@ -145,13 +145,17 @@ def password_recovery(self, request, pk=None):
...
+        if len(qs) > 1:
+            qs = qs.filter(Q(username=username) |
+                                           Q(email=username))
+
+        if len(qs) == 0:
             raise exc.WrongArguments(_("Invalid username or email"))
 
+        user = qs[0]
         user.token = str(uuid.uuid1())
         user.save(update_fields=["token"])
```
bameda: ```username``` used in line ```148``` is not defined.

Why don't you move this code to ```taiga/users/services.py```?

taiga/users/services.py
superalex: You are absolutely right, I've updated it :+1:

PR #529, opened 2015-11-17
Adding card thumbnail to attachments

superalex:


superalex:
Great!, thanks for the refactor @bameda

PR #530, opened 2015-11-18
US #2127 Discover

bameda:
https://tree.taiga.io/project/taiga/us/2127?kanban-status=38

- [x] is finished


taiga/projects/management/commands/sample_data.py
```diff
@@ -227,6 +229,13 @@ def handle(self, *args, **options):
 
             self.create_likes(project)
 
+        # Set featured projects
+        featured_projects = Project.objects.all().order_by("?")[:TOTAL_FEATURED_PROJECTS]
+        for p in featured_projects:
+            p.is_featured = True
+            p.save()
+
```
superalex: I think this can generate different featured_projects when executing the sample_data command, isn't it?. We should guarantee that if the seed is the same the result is the same too.

taiga/projects/api.py
```diff
@@ -63,19 +68,113 @@ class ProjectViewSet(LikedResourceMixin, HistoryResourceMixin, ModelCrudViewSet)
...
+        if order_by_field_name == "total_fans_last_week":
+            qs = qs.filter(likes__updated_datetime__gte=now-timedelta(days=7))
+        elif order_by_field_name == "total_fans_last_month":
+            qs = qs.filter(likes__updated_datetime__gte=now-timedelta(days=30))
+        elif order_by_field_name == "total_fans_last_year":
+            qs = qs.filter(likes__updated_datetime__gte=now-timedelta(days=365))
+        elif order_by_field_name == "total_activity_last_week":
+            qs = qs.filter(activity__updated_datetime__gte=now-timedelta(days=7))
+        elif order_by_field_name == "total_activity_last_month":
+            qs = qs.filter(activity__updated_datetime__gte=now-timedelta(days=30))
```
bameda:
```python
In [10]: from dateutil.relativedelta import relativedelta
In [11]: from django.utils import timezone
In [12]: timezone.now() - relativedelta(months=1)
Out[12]: datetime.datetime(2015, 11, 2, 10, 7, 3, 276940, tzinfo=<UTC>)
In [13]: import datetime
In [14]: datetime.datetime.strptime("2013-03-31", "%Y-%m-%d") - relativedelta(months=1)
Out[14]: datetime.datetime(2013, 2, 28, 0, 0)
In [15]: datetime.datetime.strptime("2013-12-31", "%Y-%m-%d") - relativedelta(months=1)
Out[15]: datetime.datetime(2013, 11, 30, 0, 0)
```

PR #531, opened 2015-11-19
Task: #3538: Add tribe_git json_field attribute to UserStory model

superalex:


bameda:
https://github.com/taigaio/taiga-back/pull/536

PR #535, opened 2015-11-25
Protected User Story CSV export assignment against undefined property

allistera:
I recently ran into an 500 server error when attempting to export user stories to CSV format. 

It turns out that a specific user story was created by an external user and had no owner.

In this MR I have added checking to see if the owner is set before accessing its properties.


bameda:
Merged: https://github.com/taigaio/taiga-back/commit/ed8ebfa35417a1886bae68397c750294511595e0

Thank you very much @allistera

PR #562, opened 2016-01-05
Remove usage of django.core.files.storage.Storage.path

### LINKED ISSUE #554: Support custom storage backends like django-storages (S3, Azure, CloudFiles, etc)
I am using [Deis](http://deis.io/) to host my own Taiga installation. I am running it all on Amazon Web Services, and using S3 for file storage. For my other Django projects I use django-storages' S3BotoStorage as the storage backend. However, when I enabled it in the taiga-back project I ran into a few problem areas. 

When deleting avatars and when creating a project export I would receive the following error.

`NotImplementedError: This backend doesn't support absolute paths.`

There were 3 places in the code that were using [Storage.path](https://docs.djangoproject.com/en/1.8/ref/files/storage/#django.core.files.storage.Storage.path) which raises NotImplementedError for non-local storage systems. 

I have fixed these files in my local fork. Only 3 lines needed changed that I have found so far. I'd like to discuss contributing back these changes to the main project, but I am hesitant to open a Pull Request because while the changes work on my system, I'm not sure how they would affect installations using the default local filesystem.

jeffbaier:
When using an alternate default storage (like S3 with django-storages) `django.core.files.storage.Storage.path` raises NotImplementedError. Fixes #554

These changes allow me to use S3 as my file storage with my install of Taiga. Hopefully these changes don't break users that are using the default `django.core.files.storage.FileSystemStorage`. I couldn't think of a reason to use the full filesystem path, but perhaps I'm missing something.

### DIFF
--- a/taiga/export_import/api.py
+++ b/taiga/export_import/api.py
@@ -68,8 +68,7 @@ class ProjectExporterViewSet(mixins.ImportThrottlingPolicyMixin, GenericViewSet)
             return response.Accepted({"export_id": task.id})
 
         path = "exports/{}/{}-{}.json".format(project.pk, project.slug, uuid.uuid4().hex)
-        storage_path = default_storage.path(path)
-        with default_storage.open(storage_path, mode="w") as outfile:
+        with default_storage.open(path, mode="w") as outfile:
             service.render_project(project, outfile)
 
         response_data = {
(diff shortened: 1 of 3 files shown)


bameda:
@jeffbaier tests are broken

PR #566, opened 2016-01-08
Issue 3543 - Webhooks: problem extracting the story points from a user story

superalex:
The problem was related to the use of Celery. The task was launched before the database had fully commited the update on the m2m relations of the object, now the task is launched when the database transaction is finished.


bameda:
Merged and barkported to stable branch

PR #571, opened 2016-01-11
Chinese Translation

Thomas-23:


superalex:
Hi @Thomas-23 

Thank you very much for your PR we use Transifex for Taiga translations to generate automatically all the locales files.

https://www.transifex.com/taiga-agile-llc/public/

You set up an account in Transifex and start contributing if you want.

PR #573, opened 2016-01-15
Issue 3641: taiga-back joins milestone task change and the next task change

superalex:


superalex:
@bameda I would like to speak about this PR before merging ;)

PR #580, opened 2016-01-23
make directory absolute

bclermont:
`name` value passed to this https://github.com/taigaio/taiga-back/blob/master/taiga/base/storage.py#L35 is relative, like `attachments/b/b/3/2/77785c958981dddcf2dbc013e7c27b791c7a5cab76d7a45315fc27b73b3d/btest.png`

If the running process (in this case a uwsgi process) current working isn't the value of `settings.MEDIA_ROOT` that function raise `OSError` as it couldn't create a sub directory `attachments`.

to make things safe, better to have `dirname` absolute.


superalex:
Thank you very much @bclermont 

I've just merged it here: https://github.com/taigaio/taiga-back/pull/631

PR #589, opened 2016-02-02
Max projects per user

jespino:


taiga/export_import/api.py
```diff
@@ -90,6 +91,10 @@ def create(self, request, *args, **kwargs):
         data = request.DATA.copy()
         data['owner'] = data.get('owner', request.user.email)
 
+        is_private = data.get('is_private', False)
+        if not users_service.has_available_slot_for_project(self.request.user, project=Project(is_private=is_private, id=None)):
+            raise exc.BadRequest(_("The user can't have more projects of this type"))
```
bameda: Change error message to "You can't have more private projects."

taiga/export_import/api.py
```diff
@@ -106,6 +111,9 @@ def create(self, request, *args, **kwargs):
 
         # Create memberships
         if "memberships" in data:
+            members = len(data['memberships'])
+            if not users_service.has_available_slot_for_project(self.request.user, project=Project(is_private=is_private, id=None), members=max(members, 1)):
+                raise exc.BadRequest(_("The user can't have more projects of this type"))
```
bameda: Change error message to "This project exceeds the number of members allowed."

taiga/export_import/api.py
```diff
@@ -202,17 +210,23 @@ def load_dump(self, request):
...
         except Exception:
             raise exc.WrongArguments(_("Invalid dump format"))
 
         if Project.objects.filter(slug=dump['slug']).exists():
             del dump['slug']
 
+        user = request.user
+        members = len(dump.get("memberships", []))
+        if not users_service.has_available_slot_for_project(user, project=Project(is_private=is_private, id=None), members=max(members, 1)):
+            raise exc.BadRequest(_("The user can't have more projects of this type"))
```
bameda: Here we have to now what is the error:
- "This project exceeds the number of members allowed."
- "You can't have more private projects."

taiga/export_import/management/commands/load_dump.py
```diff
@@ -58,7 +58,9 @@ def handle(self, *args, **options):
                     except Project.DoesNotExist:
                         pass
                     signals.post_delete.receivers = receivers_back
-                dict_to_project(data, args[1])
+
+                user = User.objects.get(email=args[1])
+                dict_to_project(data, user)
         except TaigaImportError as e:
             print("ERROR:", end=" ")
             print(e.message)
```
bameda: ¿? It'is not yours, I know, but If we apply the BS rule...

taiga/export_import/service.py
```diff
@@ -277,6 +277,8 @@ def store_membership(project, membership):
         serialized.save()
         return serialized
 
+    print(1111111111, serialized.errors)
+
```
bameda: :bomb:

taiga/projects/api.py
```diff
@@ -577,6 +585,10 @@ def pre_delete(self, obj):
             raise exc.BadRequest(_("The project must have an owner and at least one of the users must be an active admin"))
 
     def pre_save(self, obj):
+        if not obj.id:
+            if not users_service.has_available_slot_for_project(self.request.user, project=obj.project, members=1):
+                raise exc.BadRequest(_("The user can't have more memberships on private projects"))
+
         if not obj.token:
```
bameda: Same error here, I need to know the concrete error.

taiga/export_import/api.py
superalex: The problem can be the user can have more public projects too, it has to cover both situations.

taiga/export_import/api.py
superalex: I agree :)

taiga/export_import/api.py
superalex: Why?, if you are importing a project you should be responsible of checking you have enough project slots and memberships before trying.

taiga/export_import/service.py
superalex: :(

taiga/projects/api.py
superalex: Same comment than  before :)

taiga/users/migrations/0016_user_max_memberships_for_projects.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('users', '0015_auto_20160120_1409'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='user',
+            name='max_memberships_for_projects',
+            field=models.IntegerField(default=None, null=True, verbose_name='max number of memberships for all owned project', blank=True),
```
bameda: I don't know if you have to use here None or the settings value. What do you think?

taiga/users/migrations/0015_auto_20160120_1409.py
```diff
@@ -0,0 +1,24 @@
...
+
+    dependencies = [
+        ('users', '0014_auto_20151005_1357'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='user',
+            name='max_private_projects',
+            field=models.IntegerField(null=True, verbose_name='max number of private projects owned', default=None, blank=True),
```
bameda: I don't know if you have to use here None or the settings value. What do you think?

taiga/users/migrations/0015_auto_20160120_1409.py
```diff
@@ -0,0 +1,24 @@
...
+        ),
+        migrations.AddField(
+            model_name='user',
+            name='max_public_projects',
+            field=models.IntegerField(null=True, verbose_name='max number of public projects owned', default=None, blank=True),
```
bameda: I don't know if you have to use here None or the settings value. What do you think?

taiga/users/models.py
```diff
@@ -140,6 +141,18 @@ class User(AbstractBaseUser, PermissionsMixin):
     new_email = models.EmailField(_('new email address'), null=True, blank=True)
 
     is_system = models.BooleanField(null=False, blank=False, default=False)
+
+
+    max_private_projects = models.IntegerField(null=True, blank=True,
+                                               default=settings.MAX_PRIVATE_PROJECTS_PER_USER,
+                                               verbose_name='max number of private projects owned')
```
bameda: Use gettext in verbose_name

taiga/users/services.py
```diff
@@ -572,3 +572,33 @@ def get_voted_list(for_user, from_user, type=None, q=None):
...
+    return False
+
+
+def _has_available_slot_for_project_members(user, project, members):
+    current_memberships = project.memberships.count()
+
+    if user.max_memberships_for_projects is None:
+        return True
+    elif current_memberships + members <= user.max_memberships_for_projects:
+        return True
```
bameda:
```
def _has_available_slot_for_project_members(user, project, members):
    current_memberships = project.memberships.count()
    return ((user.max_memberships_for_projects is None) or
                 (current_memberships + members <= user.max_memberships_for_projects)
```

taiga/users/services.py
```diff
@@ -572,3 +572,33 @@ def get_voted_list(for_user, from_user, type=None, q=None):
...
+            return True
+        elif user.owned_projects.filter(is_private=True).exclude(id=project.id).count() < user.max_private_projects:
+            return True
+    else:
+        if user.max_public_projects is None:
+            return True
+        elif user.owned_projects.filter(is_private=False).exclude(id=project.id).count() < user.max_public_projects:
+            return True
+
+    return False
```
bameda: def _has_available_slot_for_project_type(user, project):
    if project.is_private:
        return ((user.max_private_projects is None) or
                     (user.owned_projects.filter(is_private=True).exclude(id=project.id).count() < user.max_private_projects))
    else:
        return ((user.max_public_projects is None) or 
                     (user.owned_projects.filter(is_private=False).exclude(id=project.id).count() < user.max_public_projects))
```

File: taiga/export_import/api.py (hunk 3 above)
Comment (bameda) [inline taiga/export_import/api.py]: Because I need to know how can I resolve the problem

File: taiga/projects/api.py [hunk 13]
```diff
@@ -550,6 +554,10 @@ def bulk_create(self, request, **kwargs):
         # TODO: this should be moved to main exception handler instead
         # of handling explicit exception catchin here.
 
+        if "bulk_memberships" in data and isinstance(data["bulk_memberships"], list):
+            if not users_service.has_available_slot_for_project(request.user, project=project, members=len(data["bulk_memberships"])):
+                raise exc.BadRequest(_("The user can't have more memberships on private projects"))
+
```
bameda: No nesting would be better

taiga/projects/api.py
```diff
@@ -577,6 +585,10 @@ def pre_delete(self, obj):
             raise exc.BadRequest(_("The project must have an owner and at least one of the users must be an active admin"))
 
     def pre_save(self, obj):
+        if not obj.id:
+            if not users_service.has_available_slot_for_project(self.request.user, project=obj.project, members=1):
```
bameda: No nesting would be better

taiga/export_import/api.py
bameda: Remember that this is an API and there is some no-human on the other side. For example some tool to import from Redmine, or trello.

taiga/export_import/api.py
bameda: So.... change error message to "You can't have more public projects."

bameda:
This PR should update the CHANGELOG too.

bameda:
Tests are broken.

PR #593, opened 2016-02-02
Fix #3846: Exclude private projects in discover calls

bameda:
Related to https://github.com/taigaio/taiga-front/pull/851


superalex:
Great!

PR #600, opened 2016-02-04
Fixing serializers validation for duplicated names in priorities, sev…

superalex:
…erities and issue types


jespino:
The use of source and "name" in different places looks inconsistent (you can use source o "name" for all the access to attrs dict).

jespino:
Looks like reapeated code, maybe moved to a Serializer Mixin for no duplication of name.

PR #601, opened 2016-02-05
Improving CSV reports performance

superalex:


taiga/projects/userstories/services.py
```diff
@@ -143,9 +144,28 @@ def userstories_to_csv(project,queryset):
                    "tags",
                    "watchers", "voters"]
 
-    for custom_attr in project.userstorycustomattributes.all():
+    custom_attrs = project.userstorycustomattributes.all()
+    for custom_attr in custom_attrs:
         fieldnames.append(custom_attr.name)
 
+    roles = project.roles.filter(computable=True).order_by('name')
```
jespino: I know was wrong in the original version, but if you want to get the csv result ordered, you must order by slug, which is the field that is used to show the points in the result.

jespino:
Looks great, I going to test it manually and if everythings works correctly and you fix my comment I will merge it :+1:

jespino:
Ready! :+1:

bameda:
And backported??

jespino:
Nope

PR #608, opened 2016-02-09
Search also into tags fields

### LINKED ISSUE #276: Search for a tag does not find the tagged US
I added a tag "node-info" to a US, now I search for node-info and does not get any result...

I would expect that tags are searched to.

volans-:
Changed the search queries to search also into the tags fields.
Fixes #276


volans-:
Tested on Taiga 1.10.0, PostgreSQL 9.4.5-0+deb8u1

bameda:
This needs performance tests with  a big DB.

bameda:
But we can make something like we do in `taiga.projects.filters.QFilterBackend` te get a rank because subject is more important than tags and tags are more important than description.

superalex:
I've tested with a big DB and the performance is ok.

I'm missing including it too in the issues and backlog sections

jespino:
Moved to the PR #784 and included some extra cnahges :smiley:

PR #609, opened 2016-02-09
Status and default status can be None

superalex:


bameda:
I remove a print in taiga/projects/history/freeze_impl.py 

M+B

PR #615, opened 2016-02-10
Fixing update_projects_order_in_bulk to check the memberships exist

superalex:
I can:
 - List my projects
 - Open a new tab and leave the team
 - Reorder from the previous tab

In this situation the API call fails because it tries to access a non existing membership.


bameda:
M+B

PR #619, opened 2016-02-12
Fixing deadlock and condition race errrors

superalex:


superalex:
@bameda @jespino we have to talk...

taiga/base/api/mixins.py
```diff
@@ -155,18 +157,39 @@ def retrieve(self, request, *args, **kwargs):
...
 
 
+def model_pk_lock(func):
+    """
+    This decorator is designed to be used in ModelViewsets methods to lock them based
+    on the model and the id of the selected object.
+    """
+    def pk_lock_wrapper(self, *args, **kwargs):
+        from taiga.base.utils.db import get_typename_for_model_class
+        lookup_url_kwarg = self.lookup_url_kwarg or self.lookup_field
```
bameda: This line has no sense. You don't use `lookup_url_kwarg`

taiga/base/api/mixins.py
```diff
@@ -155,18 +157,39 @@ def retrieve(self, request, *args, **kwargs):
...
+    @model_pk_lock
     def update(self, request, *args, **kwargs):
         partial = kwargs.pop('partial', False)
         self.object = self.get_object_or_none()
-
         self.check_permissions(request, 'update', self.object)
 
+        if self.object is None:
+            raise Http404
+
```
bameda: Can you move this one line up??? To check permissions after check if the object exist.

taiga/base/utils/db.py
```diff
@@ -126,7 +127,10 @@ def update_in_bulk_with_ids(ids, list_of_new_values, model):
     :param model: Model of the ids.
     """
     for id, new_values in zip(ids, list_of_new_values):
-        model.objects.filter(id=id).update(**new_values)
+        tn = get_typename_for_model_class(model)
```
bameda: Can this line `tn = ...` live outside the for?

taiga/base/api/mixins.py
superalex: I could but in the original update method based on DRF it's in that order too so I thought it was better maintaining the consistency

taiga/base/api/mixins.py
superalex: You are absolutely right, line removed :+1:

taiga/base/utils/db.py
superalex: Sure! :+1:

taiga/base/api/mixins.py
bameda: ok

PR #621, opened 2016-02-13
Removed 'load url from future' for compatibility with Django 1.9

AmiZya:
Since taiga-back is based on Django 1.9, it generates TemplateSyntaxError error: 'url' is not a valid tag or filter in tag library 'future' which is loaded implicitly in 1.9.


bameda:
Thanks @AmiZya

PR #627, opened 2016-02-19
usage of  cached_property

gam-phon:


superalex:
Thank you very much @gam-phon, 

I think that if using a cached_property here we should rename get_cached_user_stories to cached_user_stories in the milestone model too, this way it will be more consistent. 

@bameda have you got any comments about this?

bameda:
I agree

gam-phon:
Welcome @superalex

superalex:
Merged here: https://github.com/taigaio/taiga-back/pull/650

Thank you very much @gam-phon :+1:

gam-phon:
Welcome ^_^

PR #629, opened 2016-02-21
Fix missing import of Http404

luyikei:
https://github.com/taigaio/taiga-back/blob/master/taiga/projects/api.py#L431

There was no import of Http404 though there was the usage of it, which causes an import error.


luyikei:
I'm using Django 1.8.9

superalex:
Merged here: https://github.com/taigaio/taiga-back/pull/651

Thank you very much! :+1:

PR #630, opened 2016-02-21
Avoid accessing an attribute of a NoneType object

luyikei:
https://github.com/taigaio/taiga-back/blob/master/taiga/projects/notifications/mixins.py#L248

`obj.is_watcher = user.id in obj.watcher`

Obviously, this code will raise an error when "obj" is a NoneType object.


superalex:
Thank you very much @luyikei 

Merging here: https://github.com/taigaio/taiga-back/pull/652 :+1:

PR #635, opened 2016-02-25
Don't use the user's username in the project's slug. TG-3908

juliomenendez:
The existing code already ensured the slug yield zero collisions by querying the model for existing projects with the same slug and adding an incremental number to the new slug if a match is found. I searched the frontend repository but couldn't find any where this change will require changes in that repository.

If I'm missing something please let me know.

CC @EstherMoreno

### DIFF
--- a/taiga/projects/models.py
+++ b/taiga/projects/models.py
@@ -296,7 +296,7 @@ class Project(ProjectDefaults, TaggedMixin, models.Model):
             self.modified_date = timezone.now()
 
         if not self.slug:
-            base_name = "{}-{}".format(self.owner.username, self.name)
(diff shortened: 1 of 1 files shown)


migonzalvar:
We have been doing some housecleaning work prior to Taiga6 upcoming release and we are rejecting old PRs that seem to have been addressed already by the core team.

We believe the current Taiga release enjoys an equivalent fix and acepting this PR would cause some issues.  You could change the value of `DEFAULT_PROJECT_SLUG_PREFIX` to `False` in settings to remove the username from project slug.


Thank you very much for your contribution, it certainly inspired the team to (hopefully) find a solution to your need.

PR #658, opened 2016-03-13
Replaced references to users.User with get_user_model()

signalkraft:
While playing with an integration of the latest stable tag of Taiga into an existing Django application, I noticed that I couldn't set my own `AUTH_USER_MODEL` without breaking everything, due to hardcoded references to `users.User`. So I replaced it with `get_user_model()` and `settings.AUTH_USER_MODEL`.

I had to shuffle some `__init__` code around to get everything working without `AppRegistryNotReady` errors.


settings/testing.py
```diff
@@ -29,6 +29,17 @@
...
+    'default': {
+        'ENGINE': 'transaction_hooks.backends.postgresql_psycopg2',
+        'NAME': 'test',
+        'USER': 'postgres',
+        'PASSWORD': 'postgres',
+        'HOST': 'localhost',
+        'PORT': '',
+    }
+}
+
```
bameda: In master we use Django 1.9 and this backend is not necessary https://docs.djangoproject.com/es/1.9/releases/1.9/#s-performing-actions-after-a-transaction-commit

bameda:
I like the PR but could you fix this conflicts?

settings/testing.py
signalkraft: That was a mistake, I was testing on 1.8 and forgot to remove it. I'll do that and fix the conflicts. Should I target stable or master?

bameda: master is better for include in the next release (early April)

signalkraft:
I resolved the conflicts.

bameda:
Great

taiga/timeline/api.py
```diff
@@ -99,7 +100,7 @@ def retrieve(self, request, pk):
 
 
 class ProfileTimeline(TimelineViewSet):
-    content_type = "users.user"
+    content_type = settings.AUTH_USER_MODEL
     permission_classes = (permissions.UserTimelinePermission,)
```
bameda: tests are broken because:

```
settings.AUTH_USER_MODEL == "users.User" != "users.user"
```

bameda:
Follow at https://github.com/taigaio/taiga-back/pull/679

PR #660, opened 2016-03-15
Adding transfer_validate_token endpoint to projects API

superalex:


taiga/projects/services/transfer.py
```diff
@@ -67,12 +67,13 @@ def _validate_token(token, project_token, user_id):
     except signing.BadSignature:
         raise exc.WrongArguments(_("Token is invalid"))
 
-    if str(value) != str(user_id):
+    if str(value) != str(user.id):
         raise exc.WrongArguments(_("Token is invalid"))
 
+    print(value)
```
jespino: ups :wink:

taiga/projects/permissions.py
```diff
@@ -77,6 +77,7 @@ class ProjectPermission(TaigaResourcePermission):
     unwatch_perms = IsAuthenticated() & HasProjectPerm('view_project')
     create_template_perms = IsSuperUser()
     leave_perms = CanLeaveProject()
+    transfer_perms = IsProjectAdmin()
```
jespino: Is this correct? I think must be transfer_validate_token_perms, isn't it?

PR #679, opened 2016-03-29
Refactored occurrences of users.User to get_user_model(), which allowsa custom user model

bameda:
Related to https://github.com/taigaio/taiga-back/pull/658


bameda:
Thanks @signalkraft

PR #680, opened 2016-03-30
Fix issue #4032: Wrong validation when change the project privacity

bameda:


bameda:
I add two new test to check limints when import projects and the user is a member too.

taiga/export_import/api.py
```diff
@@ -115,11 +115,11 @@ def create(self, request, *args, **kwargs):
...
         if "memberships" in data:
-            members = len(data['memberships'])
+            members = len([m for m in data.get("memberships", []) if m["email"] != data["owner"]])
             (enough_slots, not_enough_slots_error) = users_service.has_available_slot_for_project(
                 self.request.user,
-                project=Project(is_private=is_private, id=None),
-                members=max(members, 1)
+                Project(is_private=is_private, id=None),
+                members
             )
```
superalex: Will this be considering the situation when I try to import a project where?:
 - I'm not a member
 - I'm not the owner
 - The actual number of project memberships is just the same that my actual max for that kind of project

In this situation the members param will be 0

superalex: Ignore my comment, I've just checked it :)

When importing in this point the actual number of memberships is 0 but in _has_available_slot_for_project_members the current_memberships is calculated as max(project.memberships.count(), 1) so it will always consider the importing user and the new_members is correctly filled.

Everything looks right to me :+1:

bameda: Yes, this is because this was wrong, because you were adding the number of current member twice (`current_member + members`).

PR #689, opened 2016-04-05
Clean user role permissions: remove vote_issues in Role.permissions

bameda:
Bakport in Saturdayor  Sunday


taiga/users/migrations/0018_remove_vote_issues_in_roles_permissions_field.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('users', '0017_auto_20160208_1751'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            "UPDATE users_role SET permissions = ARRAY_REMOVE(permissions, 'vote_issues')",
+            reverse_sql="UPDATE users_role SET permissions = ARRAY_APPEND(permissions, 'vote_issues')"
+        ),
```
superalex: David, won't this reverse_sql add the vote_issues role to any users_role? (not only the ones including it before the migration).

I think this migration shouldn't have a reverse_sql.

bameda: You are right

bameda:
Cool!

PR #691, opened 2016-04-05
Fix issue #4061: Invalid error when try to add more memberships

bameda:


taiga/projects/services/members.py
```diff
@@ -120,3 +120,19 @@ def check_if_project_privacity_can_be_changed(project):
...
+    if project.is_private:
+        total_memberships = project.memberships.count() + total_new_memberships
+        max_memberships = project.owner.max_memberships_public_projects
+        error_members_exceeded = ERROR_MAX_PUBLIC_PROJECTS_MEMBERSHIPS
+    else:
+        total_memberships = project.memberships.count() + total_new_memberships
+        max_memberships = project.owner.max_memberships_private_projects
+        error_members_exceeded = ERROR_MAX_PRIVATE_PROJECTS_MEMBERSHIPS
+
+    if max_memberships is not None and total_memberships_memberships > max_memberships:
```
superalex: I think it should be total_memberships and not total_memberships_memberships

PR #705, opened 2016-04-20
Adding is_out_of_owner_limits to project detail serializer

superalex:


bameda:
I've added missed tests

PR #723, opened 2016-05-11
Improving performance for history api

superalex:


taiga/projects/history/api.py
```diff
@@ -104,6 +105,12 @@ def retrieve(self, request, pk):
         self.check_permissions(request, "retrieve", obj)
 
         qs = services.get_history_queryset_by_model_instance(obj)
+        user_ids = [u["pk"] for u in qs.values_list("user", flat=True)]
+        users = get_user_model().objects.filter(id__in=user_ids)
+        users_by_id = {u.id: u for u in users}
+        for history_entry in  qs:
+            history_entry.owner = users_by_id.get(history_entry.user["pk"], None)
+
```
bameda: I don't like this a lot and the model object just have a cached property owner so you are overwriting the behavior here.

However I move this to a service (```def preload_owners_in_history_queryset(queryset)```).

superalex: Yes, the problem with that cached property is that it always make an extra query to the user data, I really neither don't like it.

I think I just had another idea about how to implement this better, let me try again ;)

superalex:
I've mixed your idea of having a preload_owners_in_history_queryset in the service with a change in the owner @property with the model and a new called prefetch_owner. I think this way is much more clear ;)

bameda:
Great!!! :sunflower: :sunglasses:

PR #734, opened 2016-05-26
Generating application token only when it doesn't exist

superalex:
If we always generate a new token when calling to the validate API we can't match the users if for example the email fails. We should talk about this with Andrés too before merging.


jespino:
:+1:

PR #749, opened 2016-06-04
Fix repetition

andela-ijubril:
Remove class that occur twice in code


bameda:
Thanks @andela-ijubril

bameda:
Merged

PR #770, opened 2016-06-24
[WIP] US #1563: Epics

bameda:
Related to https://github.com/taigaio/taiga-front/pull/1059


allan-simon:
my 2 cents here, maybe my points are not valid because the way you work are different (so by this comment I'm more looking for a return of experience), but would have it been "better" , to have this long running branch, to be splitted into smaller, feature-gated branched ? with the feature disabled by default (as not 'complete' ) , as it would have permitted to integrate faster code to master and reducing the chance of conflict at every rebase (conflicts which decrease the moral of developers and decrease the likeliness of the branch to be one day merged, at least that's what we observed on our team so far)

PR #779, opened 2016-07-07
Including the gravatar id in the responses and not generating the gra…

superalex:
…vatar avatar url by default


jespino:
Looks great :+1:

PR #782, opened 2016-07-08
Refactor Github/Gitlab/Bitbucket hooks

jespino:


jespino:
We have to review the comments added on webhook actions to be as good as possible.

PR #792, opened 2016-07-20
Fixing membership serializer

superalex:


taiga/projects/serializers.py
```diff
@@ -156,6 +156,9 @@ def get_photo(self, obj):
         return get_user_photo_url(obj.user)
 
     def get_gravatar_id(self, obj):
+        if obj.user is None:
+            return get_gravatar_id(obj.email)
```
bameda: I think this must be None because a membership with no user is an invitation and we don't know if this email will be the email of the real user.

superalex: Talking about this with @juanfran we thought it made sense, if finally the user sets another email the image can change but I don't think it's a problem.

bameda: But when you show an invitation in the admin panel you have to use no image because is an invitation... we use the same image like an unassigned item.

PR #798, opened 2016-07-28
Fixing custom fields deletes

superalex:


taiga/projects/custom_attributes/models.py
```diff
@@ -101,6 +101,7 @@ class UserStoryCustomAttributesValues(AbstractCustomAttributesValues):
     class Meta(AbstractCustomAttributesValues.Meta):
         verbose_name = "user story ustom attributes values"
         verbose_name_plural = "user story custom attributes values"
+        index_together = [("user_story",)]
```
bameda: You can use db_index in the field instead of index_together

superalex: You can but django ignores it :(

PR #816, opened 2016-08-31
Systemd compatibility scripts

DanFoxDavies:
I really don't know whether these are in the right folder or the first thing about getting the install process of them automated. This is just me acting on behalf of a colleague, ensuring that something which helped us get Taiga working on Systemd is contributed. Please feel free to move, prod and poke as required.


er1z:
In order to get this automated, the ``127.0.0.1`` IP should be used for binding, not yours, internal one.

Anyway, thanks, you made my env setting up simpler. (:

migonzalvar:
It is now documented in https://github.com/taigaio/taiga-doc/blob/89ac8d5259add76466c9032ebac5776e423c3ae8/setup-production.adoc#71-systemd-and-gunicorn

Thank you @DanFoxDavies

PR #839, opened 2016-10-05
Fixing webhooks test API call if the response isnt a valid json

superalex:


superalex:
With response.text instead of response.content we don't have to worry about encodings -> http://docs.python-requests.org/en/master/user/quickstart/#response-content

PR #843, opened 2016-10-09
simplified how the get_total_points is calculated

allan-simon:


bameda:
Merged, thank you very much @allan-simon

bameda:
https://github.com/taigaio/taiga-back/commit/ea0f496c9b7177e2efd2b2d4aa0db1299fa10055

PR #857, opened 2016-10-19
Update README.md

stadia:
it is wrong link for translate


bameda:
Rebased, thanks @stadia

PR #872, opened 2016-11-06
Add project extra info to issues, tasks and milestones list serializers

mjrk:
Returning the project extra info for issues, tasks and milestones like already done for user stories.

Reasoning: 
* without the project details, no view (name) or link (slug) for the related project can be created for a search issues, tasks or milestones search over all projects.
* it is consistent with the user stories API

If this is not desired, please give a hint for a preferred implementation (e.g. special param or a different API endpoint to use a different serializer).


superalex:
Thank you!, we like the approach but to make it more consistent we think epics and wikipages should be updated too

mjrk:
Why not? updated.

bameda:
(manual) Merged. Thanks :dancing_men:

PR #874, opened 2016-11-08
Add estimated_start, estimated_finish filters to tasks, userstories, milestones

mjrk:
Like this I can query all user stories and tasks over different projects that are supposed to end or to be started at some date. 

Tests added, feedback appreciated.


bameda:
(Manual) merged, thank you very much @mjrk

PR #883, opened 2016-11-21
Issue 4768: Notify when something is unassigned

superalex:


bameda:
Cherry-picked to stable too.

PR #890, opened 2016-11-29
Extend order_by_fields

mjrk:
For extended search (and sort) capabilities, this merge requests adds order_by_fields to milestones and tasks.

Furthermore, it adds related name fields, like type__name, status__name, severity__name etc. to order_by_fields. Thus, one can sort by status "New" over multiple projects. The relations, for example, IssueStatus, already define an ordering in their Meta like "project", "order", "name", which is not desired to be changed. However, it would be really beneficial to sort based on their name over all projects.

### DIFF
--- a/taiga/projects/issues/api.py
+++ b/taiga/projects/issues/api.py
@@ -67,6 +67,7 @@ class IssueViewSet(OCCResourceMixin, VotedResourceMixin, HistoryResourceMixin, W
                      "project__slug",
                      "status__is_closed")
     order_by_fields = ("type",
+                       "project",
                        "status",
                        "severity",
                        "priority",
--- a/taiga/projects/milestones/api.py
+++ b/taiga/projects/milestones/api.py
@@ -58,6 +58,12 @@ class MilestoneViewSet(HistoryResourceMixin, WatchedResourceMixin,
         "project__slug",
         "closed"
     )
+    order_by_fields = ("project",
+                       "name",
+                       "estimated_start",
+                       "estimated_finish",
+                       "closed",
+                       "created_date")
     queryset = models.Milestone.objects.all()
 
     def create(self, request, *args, **kwargs):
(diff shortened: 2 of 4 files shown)


bameda:
- using names produce strange beahivor with default status, types, severities... if you use non english language.
- closed_points and total_points return an HTTP 500

mjrk:
Thanks for your feedback! closed_points and total_points as well as the name sorting were removed. Is it cool now?

I did some digging regarding wrong sorting for non English languages but it does not seem to have an easy fix. The only thing I can think of would be very big CASE statement translating all names in the query to be able to sort on. And I guess, the default sorting ['project', 'order', 'name'] would also be broken in unlikely cases when 'order' and 'project' are same, right?

Any suggestions welcome.

bameda:
Great. 

The good way to order by status is using the status order because users define the correct order of status (and types, severities, priorities too) for their projects, so `status` or `status__order` should be enough

mjrk:
Not sure if I understand. How can status__order help when sorting over different projects - isn't it individually set for every project and it would result in similar strange behavior?

mjrk:
@bameda Is here still anything missing? If not, can we remove the incomplete tag and how's the chance of this PR to make it into taiga?

mjrk:
@bameda @jespino this PR is not incomplete anymore, right? Or anything more I can do?

jespino:
Sorry the absence of reply, I going to test it in my machine and merge if everything is ok.

mjrk:
@jespino Oh cool, thanks!

Good to have some progress here. I'd like to develop the extended search to be a frontend plugin some day without having to maintain a separate backend fork.

If by any chance you'd have an idea how to accomplish the localized sort by a status' name over all projects (regardingless the status__order), it might be interesting for me to have a look at it.

jespino:
Isn't possible directly with the DB information. The translation information of the main statuses is stored in the locales files, which are out of the database.

One possible option is to generate a stored procedure which the data from the locales, and generate an index from the state name using the stored procedure for each locale.

The stored procedure could be something like "if this is translable status name, index the translation, else use the text directly".

Is a quite complex solution, but I think It can work.

jespino:
Thanks!

PR #901, opened 2016-12-14
Issue 4822: Error in US estimations when created from Epics

superalex:


bameda:
@superalex test are broken

superalex:
Solving them... :)

PR #902, opened 2016-12-16
Adding management command to block projects

superalex:


jespino:
Looks good, I'm not sure with the name of the command, can be something more descriptive like "block_user_projects". Can be interesting include another parameter to set the reason (maintaining the default with the most common reason to block user projects).

superalex:
Thanks for the feedback @jespino I've changed the name, included the code parameter and also did de private filter more general.

PR #906, opened 2017-01-02
Limiting max number of pending memberships

superalex:


superalex:
Thanks @jespino it should be fixed now :)

PR #925, opened 2017-01-26
Fix Issue #5910: Create validator for assign_to attr

bameda:


taiga/projects/epics/validators.py
```diff
@@ -47,6 +48,21 @@ class Meta:
         model = models.Epic
         read_only_fields = ('id', 'ref', 'created_date', 'modified_date', 'owner')
 
+    def validate_assigned_to(self, attrs, source):
```
superalex: The different validate_assigned_to methods could be refactorized in an only Mixin

taiga/projects/issues/validators.py
```diff
@@ -37,6 +41,21 @@ class Meta:
         model = models.Issue
         read_only_fields = ('id', 'ref', 'created_date', 'modified_date', 'owner')
 
+    def validate_assigned_to(self, attrs, source):
```
superalex: The different validate_assigned_to methods could be refactorized in an only Mixin

taiga/projects/tasks/validators.py
```diff
@@ -41,6 +42,21 @@ class Meta:
         model = models.Task
         read_only_fields = ('id', 'ref', 'created_date', 'modified_date', 'owner')
 
+    def validate_assigned_to(self, attrs, source):
```
superalex: The different validate_assigned_to methods could be refactorized in an only Mixin

taiga/projects/userstories/validators.py
```diff
@@ -66,6 +67,21 @@ class Meta:
         depth = 0
         read_only_fields = ('id', 'ref', 'created_date', 'modified_date', 'owner')
 
+    def validate_assigned_to(self, attrs, source):
```
superalex: The different validate_assigned_to methods could be refactorized in an only Mixin

PR #943, opened 2017-02-28
Added common throttle for all the API

jespino:


taiga/base/throttling.py
```diff
@@ -57,5 +58,135 @@ class UserRateThrottle(throttling.UserRateThrottle):
...
+
+    def has_to_finalize(self, request, response, view):
+        return False
+
+    def allow_request(self, request, view):
+        scope = self.get_scope(request)
+        ident = self.get_ident(request)
+        rates = self.get_rates(scope)
+        if ident in settings.REST_FRAMEWORK['DEFAULT_THROTTLE_WHITELIST']:
+            return True
```
superalex: In other parts of the app we support validating range of IP's, would it be complicated do it here too?

taiga/base/throttling.py
```diff
@@ -57,5 +58,135 @@ class UserRateThrottle(throttling.UserRateThrottle):
...
+            if request.method in ["POST", "PUT", "PATCH", "DELETE"]:
+                scope = "user-write"
+            else:
+                scope = "user-read"
+        else:
+            if request.method in ["POST", "PUT", "PATCH", "DELETE"]:
+                scope = "anon-write"
+            else:
+                scope = "anon-read"
+        return scope
```
superalex: I think something like this would be more clear:
```
auth_prefix = "user" if request.user.is_authenticated() else "anon"
scope_sufix = "write" if request.method in ["POST", "PUT", "PATCH", "DELETE"] else "anon"
scope = "{}-{}".format(auth_prefix, scope_sufix)
```

jespino:
I have included your proposals, and some other minor changes, if you want to see it, are in a separated commit.

taiga/base/throttling.py
bameda: scope_sufix = "write" if request.method in ["POST", "PUT", "PATCH", "DELETE"] else **"read"**

PR #986, opened 2017-04-15
Just a small grammar fix.

LouieK22:
Am I the only one who noticed this? Can someone just accept this?

lel, I did this the wrong way, right?


jespino:
Thanks!

Issue #980, opened 2017-04-09
Feature Request: Custom Project Key for commit messages


yeoji:
For the GitHub/Bitbucket/etc integrations that allow:

- Changing of status of an epic, user story, issue with the commit message
- Attaching commits in an epic, user story, or issue

Use a custom project key to match the reference instead of TG-ref

For example, if I define my project key as HELLO, it should match HELLO-ref

timberkeley:
What is the progress on this issue? This is something that would really help with our workflow. Would it be worth testing the two pull requests or should we wait until it is merged to master? Many thanks.

PR #990, opened 2017-04-30
Add custom project key feature to Git integrations

yeoji:
This pull request is for enhancement #980 

The corresponding PR for taiga-front is [#1298](https://github.com/taigaio/taiga-front/pull/1298)

It adds the ability to define and use custom project keys to match Git commit messages in the hook. The default project_key is set to 'TG'.

github_push_messages
bitbucket_push_messages
gitlabs_push_messages
gogs_push_messages

### DIFF
--- a/taiga/hooks/bitbucket/services.py
+++ b/taiga/hooks/bitbucket/services.py
@@ -40,6 +40,11 @@ def get_or_generate_config(project):
     url = get_absolute_url(url)
     url = "%s?project=%s&key=%s" % (url, project.id, g_config["secret"])
     g_config["webhooks_url"] = url
+
+    # Use the default 'TG' project key if not defined
+    if not "project_key" in g_config:
+        g_config["project_key"] = "TG"
+
     return g_config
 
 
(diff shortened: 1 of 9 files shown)


bameda:
The PR looks great but all the tests use the default prefix, could you add some tests to use a custom prefix?

yeoji:
I've added the custom key tests for all 4 hooks :)

jespino:
Without testing it manually, looks like you use TG in uppercase by default, but the comparison is lower case based.

Could be it a problem?

yeoji:
In taiga/hooks/event_hooks.py (Line 236), the project_key is transformed into lowercase using .lower(), so there shouldn't be a problem.

`project_key = project_config[self.platform_slug]["project_key"].lower()`

jwir3:
@bameda Is this scheduled to be integrated anytime soon? It would be nice if I could use this on `master`.

bameda:
Hi @yeoji ,we have been doing some housecleaning work prior to Taiga6 upcoming release and we are rejecting old PRs that we didn't correctly attend back them. Since Taiga6 bring a signfiicant UX/UI reword, we encourage you to wait and see if Taiga6 satisfies this need and propose a new PR if that's not the case. This time we will make sure to review it when it comes through. Thank you very much on behalf of the Taiga Team!

PR #994, opened 2017-05-17
Imported projects notify incorrect users in some cases

superalex:


jespino:
I miss the conversion of the "status" field too

superalex:
Fixed! :D

PR #1020, opened 2017-08-23
Add user update throttling

jespino:


superalex:
In common.py file we have a DEFAULT_THROTTLE_RATES attribute. I think it should include an user-update entry. If not it can launch an error:

`django.core.exceptions.ImproperlyConfigured: No default throttle rate set for "user-update" scope
`

superalex:
Great!, thank you! :+1:

PR #1030, opened 2017-10-04
Add missing required dependency, django_pglocks.

jabjoe:
I was going through installing and I noticed this. Very simple fix for the next person.


jespino:
This requirement already exist in the file as a git requirement (git+https://github.com/Xof/django-pglocks.git
)

PR #1043, opened 2017-11-18
Add inspect to import, remove apps from import

ramezkabbani:
Inspect was being used without being imported so I added the import. In addition apps wasn't being used so I removed it.


bameda:
Thanks @ramezkabbani

PR #1044, opened 2017-11-24
Fix #5238: Milestone burndown graph take into account empty user stories

jespino:


jespino:
This was a quick win pending to be review. Feel free to close it or merge it :)

PR #1049, opened 2017-12-07
added custom UserManager

domdinicola:
this would allow to run createsuperuser command without getting error for is_staff field

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ static
 .\#*
 .project
 .env
+.idea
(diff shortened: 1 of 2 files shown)


bameda:
Hi @domdinicola, we have been doing some housecleaning work prior to Taiga6 upcoming release and we are rejecting old PRs that seem to have been addressed already by the core team. We believe the current Taiga release enjoys an equivalent fix and accepting this PR would cause some issues. Thank you very much for your contribution, it certainly inspired the team to (hopefully) find a solution to your need.

PR #1064, opened 2018-01-17
Case insensitive exact email match when accepting invite

internetjanitor:
This small change allows new invites to be sent to invited.person@example.com and accepted by Invited.Person@example.com.

### DIFF
--- a/taiga/projects/services/invitations.py
+++ b/taiga/projects/services/invitations.py
@@ -48,6 +48,6 @@ def find_invited_user(email, default=None):
     User = apps.get_model(settings.AUTH_USER_MODEL)
 
     try:
-        return User.objects.get(email=email)
+        return User.objects.get(email__iexact=email)
     except User.DoesNotExist:
         return default


bameda:
Hi @internetjanitor, 

Some email servers (like postfix by default, for example) distinguish addresses with capital letters. According to the RFC 5321 section 2.4 (or RFC 2821)
```
  The local-part of a mailbox MUST BE treated as case sensitive.
  Therefore, SMTP implementations MUST take care to preserve the case
  of mailbox local-parts.  In particular, for some hosts, the user
  "smith" is different from the user "Smith".  However, exploiting the
  case sensitivity of mailbox local-parts impedes interoperability and
  is discouraged.  Mailbox domains follow normal DNS rules and are
  hence not case sensitive.
```

So in Taiga those two users (`invited.person@example.com` and `Invited.Person@example.com`) could exist. That is why we cannot implement this solution.

Thank you very much for your time and for your proposal.

PR #1065, opened 2018-01-19
Feature/not found in forbidden

alexhermida:
Related to task [#5240](https://tree.taiga.io/project/taiga/issue/5240)

Only apply for getting projects by slug. Initially @migonzalvar and me prefer to be conservative changing only the retrieve `by_slug`.


superalex:
I think the wip commits (16577a5, 551926e) could be squashed with c9716e3 so there is only a commit related to the queryset filter

superalex:
Thanks! :man_dancing:

PR #1077, opened 2018-02-26
TG-1961 Assign multiple users

alexhermida:


superalex:
I've done a first review and I think I'm missing some parts of the code, the assigned_to attribute is transversal to the application so a change like this should affect at least:
- The csv reports
- The json import/export 
- The importers from external tools (trello, asana, pivotal, github...)
- The history entries for an element
- External integrations (slak...)
- The assigned_to works in the same way for tasks, issues and epics
- Migrations, what happens with the current assigned_to attribute?, I think we should simulate a migration with a huge data set and check how it works. I'm also worried about performance on another queries like for example "Items I'm involved into"
- Webhooks
- Django admin 
- Notifications
- Stats

alexhermida:
The PR is still WIP, probably we must to indicate it in the title!
Nevertheless the main idea of introduce multiple `assigned_users` it's make it simple and we try to develop it in different steps. We thought that some of the things you comment may be don't apply in this first step. 

The actual requirements for this first step:
 
* Add the feature only for US
* Add an `assigned_users` attribute without removing `assigned_to`
* History entries and notifications
* Reports

We're going to review in detail the list you comment. Thanks!

superalex:
Ups!, I thought it wasn't a WIP one :D

PR #1109, opened 2018-05-21
Disallow anonymous users to see profile details

qwertyllo:
Currently anonymous users can see other user profiles, including personal data (name, surname).
This change restricts this permission so only auth users can see other user profiles.


benyanke:
While I think this could be good, it could also break some user's workflows. It seems to be a bad idea to take existing functionality and hide it.

Perhaps this should be implemented as a configuration option which can be enabled or disabled?

qwertyllo:
Now the restriction to see profile details is enabled using a settings value (PRIVATE_USER_PROFILES) and is disabled by default.

qwertyllo:
@alexhermida Is this scheduled to be integrated soon? This option is even more important now with GDPR already implemented in several European countries.
Thanks!

localalex:
Hi @qwertyllo, can you please tell me how I can add your code into my production site? I work for a school and we can't run Taiga in its current branch because it does not have your branch implemented. Do I fork the main branch then manually add your edits? But then I'll have a forked project, how do I update it?

qwertyllo:
@localalex we should wait until this pull-request is accepted by Taiga devs, or until the same functionality is developed in main branch. If you need the funcionality as soon as possible (that was my case), you can manually apply the changes detailed in "Files changed" tab. I do not recommend using and old Taiga version.

@benyanke will this funcionality be included or developed in main branch?
It's critical to comply with GDPR in several organizations.
Thanks!

benyanke:
I'm not actually a taiga maintainer or even contributor, so I'm probably the wrong person to ask.

qwertyllo:
sorry @benyanke and thanks for your response.

@migonzalvar will this funcionality be included or developed in main branch?
It's critical to comply with GDPR in several organizations.
Thanks!

migonzalvar:
Thanks @qwertyllo

PR #1110, opened 2018-05-22
Pin kombu version to match celery

migonzalvar:
See https://github.com/celery/kombu/issues/873

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,7 @@ amqp==2.1.4
 asana==0.6.5
 bleach==2.0.0
 celery==4.0.2
+kombu==4.0.2
 cryptography==1.9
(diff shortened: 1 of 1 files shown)


alexhermida:
Good!

MasterAM:
@alexhermida not having this one in the `stable` release seemed to make our celery workers crash because of a change to `kombu`: https://github.com/celery/kombu/issues/870.

I had a ton of `KeyError: 'async'` in our celery logs, and installing kombu version 4.0.2 seems to make thing run again.

alexhermida:
@MasterAM we updated stable branch yesterday. The `stable` branch is updated during the release flow. Because of that, sometimes could be behind `master` branch.

PR #1113, opened 2018-05-28
Prevent error on US with email in history owner attribute

alexhermida:


taiga/projects/history/freeze_impl.py
```diff
@@ -88,13 +88,15 @@ def _common_users_values(diff):
...
-    if "owner" in diff:
+    if "owner" in diff and isinstance(diff["owner"], int):
         users.update(diff["owner"])
     if "assigned_to" in diff:
         users.update(diff["assigned_to"])
     if "assigned_users" in diff:
         [users.update(usrs_ids) for usrs_ids in diff["assigned_users"] if
          usrs_ids]
+
+    values["users"] = {}
```
migonzalvar: What do you think about using a [conditional expression](https://docs.python.org/3/reference/expressions.html#conditional-expressions) here?

```python
values["users"] = _get_users_values(users) if users else {}
```

alexhermida: sounds good!

PR #1166, opened 2018-08-29
Upgraded lxml, added .idea to .gitignore

SamuelMarks:


alexhermida:
@SamuelMarks regarding upgrades, those are not trivial changes and we have a strong policy to update dependencies carefully and with many tests. Towards the next major release, one of the goals will be upgrading many dependencies.

The `.gitignore` file inside the project must include configurations related to the project itself. The `.idea` folder and files are related to the programmer development environment, in your case Jetbrains IDE. Anyway thanks for the interest! 😄

PR #1192, opened 2018-11-03
Added missing migration for the 'project' model

theriverman:
In the stable (and as of today in master too) branch there's a change in the taiga/project model which has not been reflected in the migrations. Documentations(production/upgrade) do not instruct the user to invoke `python manage.py makemigrations` before applying the pre-made migrations.
If this migration has been left out intentionally, please ignore this request.


alexhermida:
@theriverman We are aware of it. Currently, we have a [#4.0rc](https://github.com/taigaio/taiga-back/tree/4.0rc) branch which includes that migration. Thank you anyway! 😄

PR #1203, opened 2018-11-22
Add my_homepage attr to projects list

alexhermida:


taiga/projects/utils.py
```diff
@@ -593,6 +593,30 @@ def attach_public_projects_same_owner(queryset, user, as_field="public_projects_
     return queryset
 
 
+def attach_my_homepage(queryset, user, as_field="my_homepage_attr"):
+    """Attach a homege array to each object of the queryset.
```
migonzalvar: s/homege/homepage

Issue #1211, opened 2018-12-04
[Fresh installation] Internal Server Error: /api/v1/userstories - No module named 'fenced_code'


benesr:
Hi,
I tried to setup an new Taiga environment (4.0.1) following the official documentation. I passed all mentioned steps successfully however parts of the system do not work, e.g. I am not able to create and save any user stories. It seems that Taiga does not find the module 'fenced_code' in the virtualenvs but I haven't find any solution to this. I added parts of the error tracelog. 

[tracelog.txt](https://github.com/taigaio/taiga-back/files/2642835/tracelog.txt)

Apologies if this is something easy, I'm pretty new to the Taiga system. Thanks in advance for any help!

davidperis92:
@benesr  Add the Bug label to this issue

antgel:
I also see this with an upgrade from v3. Kind of bemused that nobody spotted this, it makes the system unusable.

davidperis92:
In file 'taiga/mdrender/service.py' delete this line:
```
def _make_extensions_list(project=None):
     return [AutolinkExtension(),
             AutomailExtension(),
             SemiSaneListExtension(),
             StrikethroughExtension(),
             WikiLinkExtension(project),
             EmojifyExtension(),
            MentionsExtension(),
            TaigaReferencesExtension(project),
            TargetBlankLinkExtension(),

            "markdown.extensions.extra", # Remove this line!

            "markdown.extensions.codehilite",
            "markdown.extensions.sane_lists",
            "markdown.extensions.toc",
             "markdown.extensions.nl2br"] 
```
With this change the tasks and userstories can be created (See my PR)

migonzalvar:
I really try to replicate this issue but I couldn't.

Migrated instances and fresh installation are working well.

This is the minimal code example that shows how should work:

```
taiga@ubuntu-xenial:~$ cd taiga-back
taiga@ubuntu-xenial:~/taiga-back$ workon taiga
(taiga) taiga@ubuntu-xenial:~/taiga-back$ python
Python 3.5.2 (default, Nov 12 2018, 13:43:14) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import logging
>>> logging.basicConfig(level=logging.DEBUG)
>>> from markdown import Markdown
>>> md = Markdown(extensions=["markdown.extensions.extra"])
DEBUG:MARKDOWN:Successfuly imported extension module "markdown.extensions.extra".
[...]

PR #1216, opened 2018-12-13
fix(issue-1211): fix error with userstories and project creation

davidperis92:
This fix solves the error with the creation of userstories, tasks and projects.


migonzalvar:
Thank you @davidperis92 

I'm commenting on #1211. I'm not able to reproduce the issue so I need more info.

In any case, I think this pull request is not the way to go. If we deactivate the _extra_ extension we lose some features as you can see in some failing tests.

Issue #1281, opened 2019-03-31
i18n: poor german translation of ready


carlos22:
Current Translaton: 

https://github.com/taigaio/taiga-back/blob/bd385ddb5c94595ee5be46d267a3a0294c1e8323/taiga/locale/de/LC_MESSAGES/django.po#L4230-L4231

Better would be `Bereit`. As `Fertig` means `done`! In the same file there is a much better translaton for ready in the context of `Ready for test`, which is totally correct.

 https://github.com/taigaio/taiga-back/blob/bd385ddb5c94595ee5be46d267a3a0294c1e8323/taiga/locale/de/LC_MESSAGES/django.po#L4246-L4247

This needs to be changed as german speakers otherwise get "done" as second state by default which is very confusing (as there is another done state).

alexhermida:
@carlos22 Taiga uses Transifex to manage the i18n files. You can contribute to the translations at https://www.transifex.com/organization/taiga-agile-llc/ Thank you!

PR #1282, opened 2019-03-31
i18n: german translation Ready -> Bereit

carlos22:
Fertig also means Done. Issue #1281


migonzalvar:
Translations are maintained in transifex. We will make this change there.

PR #1291, opened 2019-04-15
Improvement on cors prevention

zhaoyuxuan:


zhaoyuxuan:
Change Access-Control-Max-Age from 3600 to 1800 will help prevent unsafe cors prelonged caching, which can result in unauthorized cross domain access to sensitive resources

zhaoyuxuan:
Removing "Put" "Delete" as they are considered as cors unsafe methods, which could allow attacker to inject malicious code or delete resources.

zhaoyuxuan:
removed Patch as it is also considered as preflighted request

bameda:
Merged in taiga6 ( #1531 )

Thanks @zhaoyuxuan for your contribution.

PR #1304, opened 2019-05-17
jira importer fixes

zhaoyuxuan:
Fixes for Jira importers:

1. Fix the Kanban column name doesn't match after imported
2. Add fixes to prevent crushes when doing epics import
    - this happens when our epics contains a userstory from another project in jira, and the other project hasn't been imported into taiga, so taiga can't find which userstory it will link to, and the importer crushes
3. Fix the bug that close issues didn't get assigned to its old sprint when importing from jira to taiga
4. Add logic to prevent crashing when user in jira hasn't register on taiga
    - During importing, if owner of the userstories, tasks, comments, issues hasn't register in taiga or his/her email is not present in taiga database, then importer crashed as taiga can't find that user and do not know how to deal with it
5. Fix the issue that sprint end date is imported as start date 
6. Fix the issue that sprint completed date is imported as start date 
7. Fix the issue that the wrong epics status is imported due to jira's api upgrade
8. Added ability to import work log estimate from Jira to Taiga
9. Added ability to import remaining time estimate from Jira to Taiga


taiga/importers/jira/agile.py
```diff
@@ -182,32 +186,41 @@ def _import_user_stories_data(self, project_id, project, options):
...
+                    if "closedSprints" in issue["fields"]:
+                        milestone = issue['fields']["closedSprints"][0]["name"].replace("'","")
+                        milestone = project.milestones.get(name=milestone)
 
                 us = UserStory.objects.create(
                     project=project,
                     owner=owner,
                     assigned_to=assigned_to,
-                    status=project.us_statuses.get(slug=slugify(issue['fields']['status']['name'])),
+                    status=project.us_statuses.get(slug=slugify(column_name)),
```
alexhermida: @zhaoyuxuan due to `colum_name` it's only initialized inside the [if statment](https://github.com/taigaio/taiga-back/pull/1304/files#diff-b7b431b7b3309a8317512da8d995430eR202) it could be an exception here.

taiga/projects/history/models.py
```diff
@@ -161,6 +161,8 @@ def resolve_diff_value(key):
             return (key, value)
 
         def resolve_value(field, key):
+            if field not in self.values:
```
alexhermida: Maybe, since you are updated this code you could use the `get` method for retrieving the value in the dictionary:

```diff
        def resolve_value(field, key):
-            if field not in self.values:
-              return None
-            data = self.values[field]
+            data = self.values.get(field)
```

migonzalvar:
We have been doing some housecleaning work prior to Taiga6 upcoming release and we are rejecting old PRs that we didn't correctly attend back them.

Since Taiga6 brings signficants changes, we encourage you to wait and see if Taiga6 satisfies this need and propose a new PR if that's not the case. This time we will make sure to review it when it comes through. Thank you very much on behalf of the Taiga Team!

PR #1305, opened 2019-05-17
 Hide the sensitive information in admin error reporting email

zhaoyuxuan:
Currently, there are too much sensitive information in the admin error reporting email(server configuration, code traceback and so on), these information are insecure to be left on the email and should be only put in the server log file. 

A email handler is added to filter out those sensitive information and leave only the error code, where(which file) the error occurred and the error messages in the email


bameda:
Merged in taiga6 (https://github.com/taigaio/taiga-back/pull/1532)

Thanks @zhaoyuxuan for your effort

PR #1323, opened 2019-06-12
PL translation updated

luzik:
There is very small difference between polish "ready" and "done" when using "Gotowe" (Was "Gotowe" and "Gotowe!"). "Przygotowane" is more like "prepared"


alexhermida:
@luzik thank you for your contribution! However, for transation, we use Transifex. You can check how to collaborate at https://tree.taiga.io/support/frequently-asked-questions/can-i-use-taiga-in-my-language/#how-to-help-us-to-translate-taiga-in-your-language:

You only must set up your account at https://www.transifex.com/taiga-agile-llc/public/ and as for access to the selected language.

PR #1340, opened 2019-07-04
Security improvements

lmarie23:
Add control on password strength with a regex in settings. 

Add a process to send an account activation email to the user, with a feature toggle.


bameda:
Hi @lmarie23, we have been doing some housecleaning work prior to Taiga6 upcoming release and we are rejecting old PRs that we didn't correctly attend back them. 

Related to email validation, we have implemented in Taiga5 a less intrusive validation system that requires it only if you want to add members to your project.

For password, the approach should be to use the [Django Password Validator](https://docs.djangoproject.com/en/3.1/topics/auth/passwords/#module-django.contrib.auth.password_validation) system (created for Django 2.0).

With this considerations, you could propose a new PR if you want. This time we will make sure to review it when it comes through.

PR #1358, opened 2019-09-02
Fix github importer failing when 'due_on' is not set

LibreBeats:


taiga/importers/github/importer.py
```diff
@@ -218,7 +218,7 @@ def _import_project_data(self, repo, options):
                 owner=users_bindings.get(milestone.get('creator', {}).get('id', None), self._user),
                 project=project,
                 estimated_start=milestone['created_at'][:10],
-                estimated_finish=milestone['due_on'][:10],
+                estimated_finish=milestone['due_on'][:10] if milestone['due_on'] else None,
```
migonzalvar: I'm afraid `estimated_finish` has not-null constraint. We need to look for another option when the source milestone has no estimation or due date.

migonzalvar: I would say we have two options:

- two weeks (because why not?)

- `datetime.date(datetime.MAXYEAR, 12, 31)`

What do you think @LibreBeats ?

LibreBeats: I do love the second idea, thanks!

migonzalvar:
Thank you @CarlosLVar

PR #1360, opened 2019-09-10
Markdown extension to refresh URL from attached images

migonzalvar:


migonzalvar:
@hecfernandez could you add support to this feature in taiga-front?

also please @dangarbar  take a look

PR #1362, opened 2019-09-17
Filter applicationTokenView by application

dangarbar:


migonzalvar:
It'is a good idea to add the new filter param in https://github.com/taigaio/taiga-doc/blob/master/api/application-tokens/list.adoc

PR #1369, opened 2019-10-05
add new lib connection sftp

augustogpaula:
add new lib connection sftp


bameda:
Hi @augustogpaula, we have been doing some housecleaning work prior to Taiga6 upcoming release and we are rejecting old PRs. Currently Taiga works perfectly without this dependency and we would need to know the reason for its inclusion and what it intends to solve.  

Issue #1117, opened 2018-06-04
As an unauthenticated user, bookmarks to private project pages return a 404 error message instead of showing the login page


mfriedetzki:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
As an unauthenticated user bookmarks to private projectpages like https://tree.taiga.io/project/privateExampleProject/issues return an error: 

> NOT FOUND 
> Error 404. The page you are looking for no longer exists. Perhaps you can return back to TAIGA homepage and see if you can find what you are looking for.

**If the current behavior is a bug, please provide the steps to reproduce.**

1. As an authenticated user bookmark the issue page of an private project. 
2. Log out of Taiga.
3. Open bookmarked URL as an unauthenticated user.

**What is the expected behavior?**
It should show the login page for authentication.

**Is it happening in taiga.io or in your own instance?**
taiga.io and own instance since upgrade to version 3.3.3 (version 3.1.0 showed the expected behavior)

**What browser/version are you using?**
Mozilla Firefox 60 and Google Chrome 66

**Are there any console errors *(Ctrl + F12)* in red?**
No

alexhermida:
@mfriedetzki in fact this is a `feature` not a bug. Every project URL returns a NOT FOUND for not giving extra information about registered projects. This is the original Taiga issue: https://tree.taiga.io/project/taiga/issue/5240

phgmhendrix:
Aha, but this is quite a hassle for our customers.
Could this be made a configurable option?

alexhermida:
@phgmhendrix yes of course. If we receive more feedback about that feature maybe we can develop them in the future. You can use the public project 
https://tree.taiga.io/project/taiga/ to ask for new features and vote them.

phgmhendrix:
btw. when always redirecting to the login page you do not give away any info... only after login you receive 404 or the correct page. This would be the preferable way in my humble opinion.

phgmhendrix:
As stated in [https://tree.taiga.io/project/taiga/issue/4415](https://tree.taiga.io/project/taiga/issue/4415) this issue will soon be resolved.

PR #1370, opened 2019-10-10
project not found should return unauthorized if user is authenticated

WnP:
When you receive an email from taiga and click on the link provided inside this email and are not logged inside taiga then the front end return a 404 error because this API endpoint return a 404.

This PR fix that issue. Now the front-end redirect to the login page with a next URL support.


taiga/projects/api.py
```diff
@@ -150,7 +150,13 @@ def retrieve(self, request, *args, **kwargs):
...
             qs = qs.filter(flt)
 
-        self.object = get_object_or_404(qs, **kwargs)
+        try:
+            self.object = get_object_or_404(qs, **kwargs)
+        except Http404 as e:
+            if self.request.user.is_authenticated():
+                raise e
+            else:
+                return HttpResponseUnauthorized()
```
WnP: I use a try except block here because it's less expensive than a if statement.

Also it return a 401 even if the project doesn't exists, in order to prevent attacker to brute force existing private project slugs.

WnP:
Looks like #1117 is the related issue to this PR.

WnP:
related to: https://tree.taiga.io/project/taiga/issue/5240

taiga/projects/api.py
phgmhendrix: What is the status on this issue? Has it made it to the master branche?

WnP: This issue is still pending and still present on the master branche

daniel-herrero:
@phgmhendrix I think this PR should be covered by [this other PR](https://github.com/kaleidos-ventures/taiga-back/pull/2). I close this one, but thank you very much for your effort!.

PR #1371, opened 2019-10-21
Update pillow to v6.2.0 in requirements.txt

stxm:
I have tested the proposed change with python 3..7
See: https://pypi.org/project/Pillow/6.2.0/


migonzalvar:
Thank you @stxm . We had to update all packages at once in 8b693dc

Issue #1391, opened 2020-01-31
Webhooks are not returning links for User Activity.


orientor:
I am using the normal tree.taiga.io site and I have set up a webhook receiver on my desktop. Although I am receiving webhooks they don't contain permalinks to tasks, user stories,etc as was mentioned in the docs. 
Here's one of my received JSON file.

```
{
    'action': 'change',
    'type': 'userstory',
    'by': {
        'id': 404067,
        'permalink': 'https://tree.taiga.io/profile/orientor',
        'username': 'orientor',
        'full_name': 'Aditya Verma',
        'photo': None,
        'gravatar_id': 'b72a822f0f5b7cf0c159a707af5150a1'
    },
    'date': '2020-01-31T21:16:25.080Z',
    'data': {
        'custom_attributes_values': {},
        'id': 3123912,
        'ref': 30,
        'project': {
            'id': 351023,
            'permalink': 'https://tree.taiga.io/project/orientor-sd',
            'name': 'sd',
            'logo_big_url': None
        },
        'is_closed': False,
        'created_date': '2020-01-31T18:09:19.920Z',
        'modified_date': '2020-01-31T21:16:24.919Z',
        'finish_date': None,
        'due_date': None,
        'due_date_reason': '',
        'subject': 'we',
        'client_requirement': False,
        'team_requirement': False,
        'generated_from_issue': None,
        'generated_from_task': None,
        'external_reference': None,
        'tribe_gig': None,
        'watchers': [],
        'is_blocked': False,
        'blocked_note': '',
        'description': 'sawq',
        'tags': [],
        'owner': {
            'id': 404067,
            'permalink': 'https://tree.taiga.io/profile/orientor',
            'username': 'orientor',
            'full_name': 'Aditya Verma',
            'photo': None,
            'gravatar_id': 'b72a822f0f5b7cf0c159a707af5150a1'
        },
        'assigned_to': None,
        'assigned_users': [],
        'points': [{
            'role': 'UX',
            'name': '?',
            'value': None
        }, {
            'role': 'Design',
            'name': '?',
            'value': None
        }, {
            'role': 'Front',
            'name': '?',
            'value': None
        }, {
            'role': 'Back',
            'name': '?',
            'value': None
        }],
        'status': {
            'id': 2107163,
            'name': 'New',
            'slug': 'new',
            'color': '#999999',
            'is_closed': False,
            'is_archived': False
        },
        'milestone': {
            'id': 254091,
            'name': 'eres',
[...]

PR #1393, opened 2020-02-01
[issues/1391] fix: permalink serialization

theriverman:
Fixing a webhook serialization bug raised in [Issue 1391](https://github.com/taigaio/taiga-back/issues/1391) where `permalink` fields were not added to the impacted object's webhook payload.

`permalink = serializers.SerializerMethodField("get_permalink")` => `permalink = MethodField()`


theriverman:
CI failed here for Python 3.5 only (3.6 and 3.7 have both passed):
`tests/integration/test_searches.py::test_search_text_query_in_my_project FAILED`
Which is line `142` in `tests/integration/test_searches.py`:
```python
response = client.get(reverse("search-list"), {"project": data.project1.id, "text": "back"})
assert response.status_code == 200
assert response.data["count"] == 15
```
But it makes no sense to me to fail here. Could somebody have a look at it and share their thoughts, please

migonzalvar:
Sometimes test randomly fail, not sure why. We restart the job and worked like a charm.

migonzalvar:
Thank you again @theriverman

PR #1395, opened 2020-02-04
Require Django 1.11.28 to avoid SQL Injection

stxm:
This is related to CVE:
https://nvd.nist.gov/vuln/detail/CVE-2020-7471


migonzalvar:
Thank you @stxm We need to update all dependeincias at once. Now is merged in master 27a488bdad2e0e9577abe4e83778062ada9429ac

PR #1397, opened 2020-02-05
checking if slug attribute exists

Meissnerim:
getattr raises  "AttributeError: 'xxxx' object has no attribute 'slug'" if the attribute does not exist. 
See https://docs.python.org/3/library/functions.html#getattr

I propose to check with hasattr() instead.


migonzalvar:
You are right.

Instead of `hasattr` we can just add _default_

```
slug = getattr(self.project, "slug", default=None)
```

Meissnerim:
Thanks for the quick response and the fix in #1398

PR #1414, opened 2020-03-18
Some changes to allow running it with django 2.2

jespino:
Probably some changes are missing, but is a initial step.


migonzalvar:
Hi @jespino. We cherry picked and get inspiration from this branch. It is now in master branch. Thank you very much!

PR #1415, opened 2020-03-27
docs: Fix simple typo, accesible -> accessible

timgates42:
There is a small typo in tests/integration/test_users.py.

Should read `accessible` rather than `accesible`.


migonzalvar:
Thanks @timgates42

PR #1424, opened 2020-04-24
Exclude non-member users from mentions

dangarbar:


taiga/mdrender/extensions/mentions.py
```diff
@@ -30,19 +30,33 @@
...
+        super().__init__(pattern, md)
+
     def handleMatch(self, m):
         username = m.group(3)
-
         try:
-            user = get_user_model().objects.get(username=username)
+            user = get_user_model().objects.get(
+                username=username,
+                memberships__project_id=self.project.id,
```
migonzalvar, line 58: Warning, `self.project` can be `None` in some corner cases. It is important to tkae into account this case.

PR #1431, opened 2020-05-11
Update requirements.txt

benz310:


migonzalvar:
Thank you @benz310. I'm afraid we have to manage dependencies as a whole, not only in `requirements.txt`. We hope we can update this and other packages soon.

PR #1436, opened 2020-05-20
ready-for-test status is currently implying a ticket is closed. 

virgiliolino:
Greetings,
after so many years fighting in teams for a definition of ready and done I could not rapidly create this pull request.

**previous situation**
When moving a tasks to the column "ready for test", the task is considered as completed.
As a consequence for example User stories containing those tickets could be accounted as a progress, velocity is then distorted and other small impressions.

*with this fix*
At least for new installations this issue will be fixed. 

I considered not creating a migration for currently existing installations cause we would change the behavior in projects that perhaps are used on working in the wrong way.

### DIFF
--- a/taiga/projects/fixtures/initial_project_templates.json
+++ b/taiga/projects/fixtures/initial_project_templates.json
@@ -225,7 +225,7 @@
                 "color": "#ffcc00",
                 "name": "Ready for test",
                 "order": 3,
-                "is_closed": true
+                "is_closed": false
             },
             {
                 "slug": "closed",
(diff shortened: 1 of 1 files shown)


diacritica:
Hi @virgiliolino 
Thanks for your suggestions. You seem to prefer default behaviour as OPEN for Ready-for-Test template status instead of CLOSED.
We are in the process of revisiting many of these decisions during June, I will keep a pointer to your suggestion. Having said that, I feel inclined to accept your proposal.

bameda:
Thanks @virgiliolino for your proposal and your time.

PR #1437, opened 2020-05-22
upgrade psycopg2-binary

bhanuvrat:
2.7.5 crashes during installation.
2.8.5 works as mentioned here https://stackoverflow.com/a/56731492/338691


migonzalvar:
Updated in master

PR #1440, opened 2020-06-02
feat: Read Celery config from env var

madmath03:
feat: Read Celery config from env var

Default values are still the same as before, only now it's possible to initialize them through environment variables. It's particularly useful if used in a container like docker.

Signed-off-by: mathieu.brunot <mathieu.brunot@monogramm.io>


theriverman:
I think this change brings inconsistency to the overall configuration of Taiga.
Changing this config file only and leaving common, local, celery, etc intact is not good.

Instead, I recommend you to create a `celery_local.py` in your `taiga-back/settings` directory and customise it with env vars exactly as you'd like it. This file will be gitignore'd by default, no VCS trouble.

The `celery_local` file should start with a relative import to the stock celery file:
```python
from .celery import *
from os import environ

# ... your config comes here
```

This custom `celery_local.py` will be picked up by Taiga automatically:
https://github.com/taigaio/taiga-back/blob/5587fb905b15bdc91bb04eaf2eae9b372392dd4b/taiga/celery.py#L27-L30

madmath03:
Thanks @theriverman !

You're right, I completely missed that part, so I guess this PR is not needed at all.
I was following the Taiga documentation but saw no mention of that: <https://taigaio.github.io/taiga-doc/dist/setup-production.html#_async_tasks_optional>

Might be nice to update it.

theriverman:
You're welcome!
Yeah, it should probably be documented somewhere, I agree.

PR #1450, opened 2020-06-12
Tg 101 - Task promotion create user story and delete original task

hecfernandez:


taiga/projects/userstories/models.py
```diff
@@ -136,10 +136,8 @@ class UserStory(OCCModelMixin, WatchedModelMixin, BlockedMixin, TaggedMixin, Due
                                              on_delete=models.SET_NULL,
                                              related_name="generated_user_stories",
                                              verbose_name=_("generated from issue"))
-    generated_from_task = models.ForeignKey("tasks.Task", null=True, blank=True,
```
migonzalvar, line 139: We can remove the function but I think this field should be kept for backward compatibility reasons.

PR #1454, opened 2020-06-16
Tg 101 - Task promotion create user story and delete original task

hecfernandez:


taiga/projects/services/promote.py
```diff
@@ -39,7 +40,7 @@ def promote_to_us(source_obj):
     for obj in queryset:
         us = UserStory.objects.create(
             generated_from_issue_id=obj.id if model_class.__name__ == "Issue" else None,
-            generated_from_task_id=obj.id if model_class.__name__ == "Task" else None,
+            from_task_ref = _("Task #(%(ref)s)") % {"ref": obj.ref} if model_class.__name__ == "Task" else None,
```
migonzalvar, line 43: Not sure it this is OK:

This will render: `"Task #(123)"`

Should be: `"Task (#123)"`

taiga/projects/mixins/promote.py
```diff
@@ -41,4 +41,8 @@ def promote_to_user_story(self, request, pk=None):
         ret = promote_to_us(obj)
         self.persist_history_snapshot(obj=obj)
 
+        # delete source task if required
+        if obj.__class__.__name__ == "Task":
```
migonzalvar, line 45: It's more common 

```
if isinstance(obj, Task):
```

PR #1476, opened 2020-09-01
Refactor settings

yamila-moreno:
This PR has a refactor to the settings files so it'll be easier to integrate with docker and docker-compose.

The main changes are:
- `settings/testing.py` is moved to `tests/config.py` so the context of its use is clearer
- `development.py` and `local.py` were removed; we'll enforce the use of `DJANGO_SETTINGS_MODULE=my.config python manage.py runserver` 
- two new example files for _dev_ and _prod_ environments
- `settings/__init__.py` is no longer needed because CELERY settings are merged in `common` file

**To test this PR**

- `python manage.py runserver` alone should work
- `DJANGO_SETTINGS_MODULE=settings.config python manage.py runserver` should work with new settings (after creating the file)
- `py.test` should also work


migonzalvar (approved):
Great! Clearer and easier to reason about.

PR #1480, opened 2020-09-10
fix: Only shows verify button when unverified

CarlosLVar:
@migonzalvar

### DIFF
--- a/taiga/users/templates/emails/registered_user-body-html.jinja
+++ b/taiga/users/templates/emails/registered_user-body-html.jinja
@@ -3,14 +3,18 @@
 {% block body %}
 <table border="0" width="100%" cellpadding="0" cellspacing="0" class="table-body">
     <tr>
-        {% trans signature=sr("signature"), product_name=sr("product_name"), url=resolve_front_url("verify-email", user.email_token), support_email=sr("support.email") %}
+        <td>
+            {% if user.verified_email %}
+              {% trans url=resolve_front_url("verify-email", user.email_token) %}
+              <p>Please confirm your email</p>
+              <a class="button" href="{{ url }}" title="Confirm email">Confirm email</a>
+              {% endtrans %}
+            {% endif %}
+        </td>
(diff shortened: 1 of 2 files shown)


taiga/users/templates/emails/registered_user-body-html.jinja
```diff
@@ -3,12 +3,18 @@
 {% block body %}
 <table border="0" width="100%" cellpadding="0" cellspacing="0" class="table-body">
     <tr>
-        {% trans signature=sr("signature"), product_name=sr("product_name"), url=resolve_front_url("verify-email", user.email_token), support_email=sr("support.email") %}
+        <td>
+            {% if user.verified_email %}
```
migonzalvar, line 7: It is the opposite, isn't it?

             {% if not user.verified_email %}

Also we can move the if so it wraps the `<td>`

CarlosLVar: Indeed it is! I'll apply those fixes now

taiga/users/templates/emails/registered_user-body-text.jinja
```diff
@@ -1,12 +1,17 @@
-{% trans signature=sr("signature"), product_name=sr("product_name"), url=resolve_front_url("verify-email", user.email_token), support_email=sr("support.email") %}
+{% trans %}
```
migonzalvar, line 1: product_name undefined here

taiga/users/templates/emails/registered_user-body-text.jinja
```diff
@@ -1,12 +1,17 @@
...
 We're thrilled you've joined our growing community of professionals revolutionizing their way of working and hope you enjoy it.
+{% endtrans %}
 
+{% if not user.verified_email %}
+{% trans url=resolve_front_url("verify-email", user.email_token) %}
 Please confirm your email: {{ url }}
+{% endtrans %}
+{% endif %}
 
+{% trans signature=sr("signature"), product_name=sr("product_name"), support_email=sr("support.email") %}
```
migonzalvar, line 13: product_name unused

PR #1482, opened 2020-09-14
Fix/emoji processing on url

carlosgoce:
@hecfernandez @migonzalvar 

Maybe this is enough?

### DIFF
--- a/tests/unit/test_mdrender.py
+++ b/tests/unit/test_mdrender.py
@@ -242,6 +242,18 @@ def test_render_triple_quote_and_lang_code():
     assert render(dummy_project, "```python\nprint(\"test\")\n```") == expected_result
 
 
+def test_render_url():
+    source = "[Link](https://example.com/a/x/1)"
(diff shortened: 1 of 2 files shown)


tests/unit/test_mdrender.py
```diff
@@ -242,6 +242,18 @@ def test_render_triple_quote_and_lang_code():
...
 
+def test_render_url():
+    source = "[Link](https://example.com/a/x/1)"
+    expected_result = '<p><a href="https://example.com/a/x/1" target="_blank">Link</a></p>'
+    assert render(dummy_project, source) == expected_result
+
+
+def test_render_url_with_emoji_like():
+    source = "[Link](https://example.com/a/:x:/1)"
+    expected_result = '<p><a href="https://example.com/a/\:x\:/1" target="_blank">Link</a></p>'
```
carlosgoce, line 253: Notice the backslash in the URL scaping colons
 \:x\:

migonzalvar:
Not sue if it is necessary after change in taiga-front

migonzalvar:
It is not necessary now

PR #1490, opened 2020-09-22
fix: Removes <td> element from registered email

CarlosLVar:
@migonzalvar

### DIFF
--- a/taiga/users/templates/emails/registered_user-body-html.jinja
+++ b/taiga/users/templates/emails/registered_user-body-html.jinja
@@ -4,12 +4,10 @@
 <table border="0" width="100%" cellpadding="0" cellspacing="0" class="table-body">
     <tr>
         {% if not user.verified_email %}
-        <td>
               {% trans url=resolve_front_url("verify-email", user.email_token) %}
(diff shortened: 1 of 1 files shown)


taiga/users/templates/emails/registered_user-body-html.jinja
```diff
@@ -4,12 +4,10 @@
 <table border="0" width="100%" cellpadding="0" cellspacing="0" class="table-body">
     <tr>
         {% if not user.verified_email %}
-        <td>
```
migonzalvar, line 7: It seems to me it's not valid HTML

Maybe:

```jade
table
  tr
    td
      {% if not user.verified_email %}
      {% endif %}
      {% trans signature=sr("signature"), product_name=sr("product_name"), support_email=sr("support.email") %}
  tr
    td
      h2 Thank you for registering in ...
```

```

CarlosLVar: You are absolutely right, I'm closing this one and starting a new PR at https://github.com/taigaio/taiga-back/pull/1516

PR #1520, opened 2020-11-19
issue/968/Fix createsuperuser

tdelatorre:
Fix createsuperuser and fix empty email


taiga/users/models.py
```diff
@@ -91,6 +91,39 @@ def get_user_file_path(instance, filename):
...
+        Creates and saves a superuser with the given username, email, password.
+        """
+        user = self.create_user(
+            username=username,
+            email=email,
+            password=password
+        )
+        user.is_superuser = True
+        user.is_admin = True
+        user.save(using=self._db)
```
yamila-moreno, line 123: why the param "using=self._db" ?

taiga/users/models.py
```diff
@@ -25,7 +25,7 @@
 from django.apps import apps
 from django.apps.config import MODELS_MODULE_NAME
 from django.conf import settings
-from django.contrib.auth.models import UserManager, AbstractBaseUser
```
yamila-moreno, line 28: UserManager is no longer needed

taiga/users/models.py
tdelatorre: If someone has another database defined in their config.py this uses the default. Is the same as using=None but I don't think it's necessary. I remove it.

taiga/users/models.py
tdelatorre: Thanks! I remove it.

Issue #993, opened 2017-05-12
error: filter function return all user stories use "assigned to"


ban2007:
**Before adding an issue**
Please ensure that your issue is not already reported in our [issues list](https://tree.taiga.io/project/taiga/issues?order_by=-created_date).
If this issue was already reported, remember that you can upvote it to raise its importance.

**Do you want to request a *feature* or report a *bug*?**
It's a bug.
**What is the current behavior?**
I add 8 user, create a KanBan project, and add some user-stories. 
sometime, I filter a "assigned to" user B, but the final result includes all user's user-stories
**If the current behavior is a bug, please provide the steps to reproduce.**
1. open kanban, filter "assigned to" user A task, response correct.
2. quickly and successively delete this case and change another filter "assigned to" user B. It must be quickly enough.
The result shows right result only including user B user-story, but quickly change to show all user's user-story
3. it's not reproduce every time, but with high probability.

**Is it happening in taiga.io or in your own instance?**
my own instance
**What browser/version are you using?**
firefox
**Are there any console errors *(Ctrl + F12)* in red?**
no
Thanks for reporting!

Issue #991, opened 2017-05-05
Automatically move stories to next sprint


alex-hutton:
Hi, thanks for the great software.

This is a feature request.

When the timeframe of a sprint ends, I would like the option to have unclosed stories moved automatically to the next chronological sprint.

If this feature is desired I will try and submit a PR for it.

superalex:
Thank you very much @alex-hutton I think this one is duplicated here: https://tree.taiga.io/project/taiga/issue/2371

If you are interested in proposing such a major change you can follow this guide with the steps required to include a PR in taiga: https://tree.taiga.io/support/frequently-asked-questions/how-can-i-contribute-to-taiga/#propose-some-feature-or-enhancement

Regards,

PR #1525, opened 2020-11-23
feat(project) #993 add default_swimlane to project

bameda:
This PR adds default_swimlane to the model `Project` and creates a signal to update the value each time the first swimlane of a project is created. Obviously you can modify it like the rest of the project params. 

You only can delete a swimlane if it's the last or it's not the default one.

One more thing, to ensure that the API remains consistent, If the project of a user story is modified, the swimlane will be assigned by default if the kanban module of the new project is activated, else it set to None.

## How To Validate:

- [x] Prerequisites
  - [x] Regenerate data with sample data.
  - [x] Serve _taiga-front_. 
  - [x] Set `TOKEN`
    ```
    TOKEN=$(
      curl -X POST \                       
        -H "Content-Type: application/json" \
        -d '{"type": "normal", "username": "admin", "password": "123123"}' \
        http://localhost:8000/api/v1/auth | jq -r '.auth_token'
    )

    ```
- [x] Test 0: Review
  - [x] Review the code
  - [x] Review the test code to understand how it works
  - [x] Pass the tests.
- [x] Test 1: Create the first swimlane to check signal behavior.  
  - [x] Go to a project without swimlanes, for example `projects 6`.
  - [x] Check default swimlane value is `None`
    ```
     curl -v -X GET \
       -H "Content-Type: application/json" \
       "http://localhost:8000/api/v1/projects/6" | jq "." | grep default_swimlane
    ```
  - [x] Create first swimlane.
  - [x] Check default swimlane has changed
    ```
     curl -v -X GET \
       -H "Content-Type: application/json" \
       "http://localhost:8000/api/v1/projects/6" | jq "." | grep default_swimlane
    ```
  - [x] Create the second swimlane
  - [x] Check default swimlane has not changed
    ```
     curl -v -X GET \
       -H "Content-Type: application/json" \
       "http://localhost:8000/api/v1/projects/6" | jq "." | grep default_swimlane
    ```
- Test 2: delete swimlanes:
  - [x] Get default swimlane id for project 6 (for example `16`)
    ```
     curl -v -X GET \
       -H "Content-Type: application/json" \
       "http://localhost:8000/api/v1/projects/6" | jq "." | grep default_swimlane
    ```
  - [x] Get all project swimlanes ids (for example `16` and `17`)
  - [x] Try to remove default swimlane (`16`). It returns and error because is the default one.
    ```
    curl -v -X DELETE \
      -H "Content-Type: application/json" \
      -H "Authorization: Bearer $TOKEN" \
      "http://localhost:8000/api/v1/swimlanes/16?moveTo=17" | jq "."
    ```
  - [x] Try to remove not default swimlane (`17`). The response should be 204 and the swimlane should be deleted
    ```
    curl -v -X DELETE \
      -H "Content-Type: application/json" \
      -H "Authorization: Bearer $TOKEN" \
      "http://localhost:8000/api/v1/swimlanes/17?moveTo=16" | jq "."
    ```
  - [x] Try to remove default swimlane (`16`) again. Now you can remove it without problems.
    ```
    curl -v -X DELETE \
      -H "Content-Type: application/json" \
      -H "Authorization: Bearer $TOKEN" \
      "http://localhost:8000/api/v1/swimlanes/16" | jq "."
    ```
  - [x] Check default swimlane value is `None`
    ```
     curl -v -X GET \
       -H "Content-Type: application/json" \
       "http://localhost:8000/api/v1/projects/6" | jq "." | grep default_swimlane
    ```
- [x] Move [task #993](https://tree.taiga.io/project/taiganext/task/993) to _ready-for-qa_
- [x] Unblock [task #991](https://tree.taiga.io/project/taiganext/task/991)


tests/factories.py
```diff
@@ -768,6 +768,21 @@ def create_project(**kwargs):
     return project
 
 
+def create_swimlane(**kwargs):
+    "Create a swimlane and along with its dependencies."
```
yamila-moreno, line 772: what does it mean?

bameda: Fixed:      "Create a swimlane ~~and~~ along with its dependencies."

PR #1534, opened 2020-12-02
Bulk emails in docker

tdelatorre:


docker/config.py
```diff
@@ -45,13 +45,17 @@
...
 
+
 #########################################
 ## EMAIL
 #########################################
 # https://docs.djangoproject.com/en/3.1/topics/email/
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 DEFAULT_FROM_EMAIL = os.getenv('DEFAULT_FROM_EMAIL', 'system@taiga.io')
 ENABLE_EMAIL = os.getenv('ENABLE_EMAIL', False) in (True, 'True')
+CHANGE_NOTIFICATIONS_MIN_INTERVAL = 30  # seconds
```
yamila-moreno, line 56: Better 60 seconds

taiga/projects/notifications/services.py
```diff
@@ -563,3 +567,16 @@ def make_ms_thread_index(msg_id, dt):
...
+    with advisory_lock("send-notifications-command", wait=False) as acquired:
+            if acquired:
+                qs = HistoryChangeNotification.objects.all().order_by("-id")
+                for change_notification in iter_queryset(qs, itersize=100):
+                    try:
+                        send_sync_notifications(change_notification.pk)
+                    except HistoryChangeNotification.DoesNotExist:
+                        pass
+            else:
+                print("Other process already running")
```
bameda, line 582: This else statement should be removed, it has no sense in a service

docker/config.py
tdelatorre: Done!

taiga/projects/notifications/services.py
tdelatorre: Done!

PR #1550, opened 2021-01-07
(feat) add importers

yamila-moreno:
Not here! GOTO: https://github.com/taigaio/taiga-docker/pull/11


docker/config.py
```diff
@@ -133,3 +133,43 @@
...
+if ENABLE_JIRA_IMPORTER:
+    IMPORTERS["jira"] = {
+        "active": True,
+        "consumer_key": os.getenv('JIRA_IMPORTER_CONSUMER_KEY'),
+        "cert": os.getenv('JIRA_IMPORTER_CERT'),
+        "pub_cert": os.getenv('JIRA_IMPORTER_PUB_CERT')
+    }
+
+ENABLE_TRELLO_IMPORTER = os.getenv('ENABLE_TRELLO_IMPORTER', 'False') == 'True'
+if ENABLE_IMPORTER_GITHUB:
```
bameda, line 170: This should be `ENABLE_TRELLO_IMPORTER` instead of `ENABLE_IMPORTER_GITHUB`

yamila-moreno: fixed!

PR #1551, opened 2021-01-08
Fix createsuperuser

tdelatorre:
Fix a bug in createsuperuser. Now you can do: python manage.py createsuperuser


taiga/users/models.py
```diff
@@ -116,10 +115,9 @@ def create_superuser(self, username, email, password):
         user = self.create_user(
             username=username,
             email=email,
-            password=password
+            is_superuser=True
         )
-        user.is_superuser = True
-        user.is_admin = True
+        user.set_password(password)
```
yamila-moreno, line 120: set_password is already done in self.create_user

taiga/users/models.py
```diff
@@ -116,10 +115,9 @@ def create_superuser(self, username, email, password):
...
         user.save()
```
yamila-moreno, line 121: this .save is already done in create_user, should be removed from here

tdelatorre: Done!

tdelatorre: Done!

PR #1554, opened 2021-01-08
fix(admin): update kanban order after userstory status is deleted

bameda:
This PR makes sure that when you delete a status and move your user stories to another, they are placed below the ones that already exist. Just updating Kanban_order (recalculating it) for the mover stories.

- [x] To test it, just delete statuses in a project with kanban enbaled and see the behavior
- [x] Move [task #1148](https://tree.taiga.io/project/taiganext/task/1148) to ready-for-qa


taiga/projects/mixins/on_destroy.py
```diff
@@ -62,16 +67,17 @@ def destroy(self, request, *args, **kwargs):
 
         # if the object is not deleted
         if not isinstance(res, response.NoContent):
-            # Restart estatus if we can't delete the object
+            # Restart status (roolback) if we can't delete the object
```
yamila-moreno, line 70: roolback ==> rollback?

bameda: # :facepalm:

PR #1574, opened 2021-02-04
Set lowercase string comparison on ENV VAR so that it matches any…

swiss-knight:
Set lowercase string comparison on ENV VAR so that it matches any user input.

Help fixing: https://github.com/taigaio/taiga-docker/issues/23


swiss-knight:
Fire checks again.

bameda:
I close the PR because, even though the proposed changes make sense, it generates certain incompatibilities that we decide to resolve with more biggest changes in a future mayor release. 

Thank you very much for your contribution and for using Taiga, @swiss-knight 

Best regards

PR #1577, opened 2021-02-07
Change the rabbitmq host to configurable (read from ENV)

ginuerzh:
The rabbitmq host in `docker/config.py` is hardcoded (`taiga-events-rabbitmq` and `taiga-async-rabbitmq`), it's OK for docker-compose, but in k8s the name may be different, such as `rabbit.mq.svc.cluster.local`. So it would be better to be open as configurable.


yamila-moreno:
Hi @ginuerzh thanks for your contribution. In order to this change to be complete, there would be necessary other changes. It would be awesome if you'd like to make them:

- PR to taiga-docker/README.md adding the envvar in the explanation
- PR to taiga-doc/setup-production.adoc adding the envar in the explanation

Best!

yamila-moreno:
Thanks for your contribution @ginuerzh !! :)

PR #1599, opened 2021-02-24
Fix Jira import issue (user list not being filtered by selected project)

argentaiga:
**Issue:** 
Jira import process fails retrieving the list of users to be imported. 

**Cause:**
This happens when the Jira instance has a large amount of projects. The import is trying to get the list of users from all existing projects (instead of filtering by the project that is selected in the import wizard). 

**Fix:**
Code was modified to get the list of users based on the project that is selected in the import wizard.

### DIFF
--- a/taiga/importers/jira/api.py
+++ b/taiga/importers/jira/api.py
@@ -64,7 +64,7 @@ class JiraImporterViewSet(viewsets.ViewSet):
 
         importer = JiraNormalImporter(request.user, url, token)
         try:
-            users = importer.list_users()
+            users = importer.list_users(project_id)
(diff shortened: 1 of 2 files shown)


argentaiga:
@yamila-moreno @migonzalvar, we created this PR and noticed some checks are failing. I tried to run the tests in my local environment but couldn't make it work yet. I'm wondering if the test that is breaking is related to our fix, or it's just something that was already failing. Can you confirm?

migonzalvar:
@argentaiga We are still testing in our staging servers. Unfortunately something is not working as expected. I hope I can tell you more soon.

PR #1611, opened 2021-04-13
Fixes docker build on arm64 architectures.

alexmarkley:
For users who want to run Taiga on AWS Graviton or Apple Silicon, arm64 docker images are required.

This pull request fixes the build on arm64 (and probably other non-intel architectures as well) by adding the required build dependencies for pip to rebuild psychopg2 from source when there is no suitable binary available to download.

See https://gallery.ecr.aws/r2d7r3i9/taiga-back for proof of concept builds using this fix.

Related: https://github.com/taigaio/taiga-docker/issues/70

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -36,61 +36,62 @@ WORKDIR /taiga-back
 ENV GOSU_VERSION 1.12
 
 RUN set -eux; \
-    savedAptMark="$(apt-mark showmanual)"; \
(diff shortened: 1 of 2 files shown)


alexmarkley:
In the process of standing up my Taiga 6 environment, I made some additional fixes to this Dockerfile. You can see it running on my ECS cluster (arm64, not x86) here: https://taiga.mbe.tv/project/yerface/timeline

PR #1616, opened 2021-05-05
:sparkles: Add settings for new backlog options.

jwir3:
This adds two new settings for projects: default_backlog_order_scrum and
default_backlog_order_kanban. Each of these is used to control the users'
settings for whether new stories should be added to the top or bottom
(default) of the respective workflows.


jwir3:
@migonzalvar Review ping?

